PR #16, opened 2026-03-04
Fix: Add vertices field support for polygon geofences

excing:
## Problem

When adding polygon geofences with vertices parameter, the polygon data was not being persisted to the database. This caused polygon geofences to fail to display or function correctly after being saved.

## Root Cause

The Geofence model in Dart already had the vertices field defined, but the database schemas on both Android and iOS platforms were missing the corresponding column to store this data. When addGeofence() was called, the vertices data was silently dropped during database insertion.

## Solution

Android (TraceletDatabase.kt):
  - Added vertices TEXT column to the geofences table schema
  - Incremented database version from 3 to 4
  - Added migration logic in onUpgrade() to add the column for existing databases
  - Updated insertGeofence() to serialize vertices as JSON array ([[lat1,lng1],[lat2,lng2],...])
  - Updated cursorToGeofence() to deserialize vertices from JSON when reading

iOS (TraceletDatabase.swift):
  - Added vertices TEXT column to the geofences table schema
  - Updated insertGeofence() to serialize vertices as JSON array
  - Updated geofenceRowToMap() to deserialize vertices from JSON when reading

## Breaking Changes

None. This is a backward-compatible fix. Existing circular geofences continue to work as before.

### DIFF
--- a/packages/tracelet_android/android/src/main/kotlin/com/tracelet/tracelet_android/db/TraceletDatabase.kt
+++ b/packages/tracelet_android/android/src/main/kotlin/com/tracelet/tracelet_android/db/TraceletDatabase.kt
@@ -21,7 +21,7 @@ class TraceletDatabase private constructor(context: Context) :
 
     companion object {
         private const val DB_NAME = "tracelet.db"
-        private const val DB_VERSION = 3
+        private const val DB_VERSION = 4
 
         // Location table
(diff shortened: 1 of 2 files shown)


GalacticTitan:
I am respectfully closing it because of the issues mentioned.. but I am implementing this fix . Thank you.

Issue #51, opened 2026-04-13
Geofence callbacks not triggering during continuous tracking


Srj-Love:
Hi Tracelet team,

I am testing geofence support together with continuous tracking.

My setup:
- I use `Tracelet.start()` for continuous location tracking
- I register geofences using `addGeofences(...)`
- location tracking works correctly
- HTTP sync also works correctly
- but `onGeofence(...)` is not firing when the device enters the geofence radius

Important note:
- I am testing with mock locations
- `rejectMockLocations` is set to `false`
- raw location callbacks continue to arrive correctly
- only geofence callbacks are missing

I am not using `startGeofences()` because my app requires continuous vehicle tracking and geofence events in the same flow.

Expected behavior:
- while running `Tracelet.start()`
- after registering geofences with `addGeofences(...)`
- entering or exiting a geofence should trigger `onGeofence(...)`

Actual behavior:
- `onLocation(...)` works
- `onHttp(...)` works
- `onGeofence(...)` does not fire

Here is a minimal sanitized repro:

```dart

import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:geocoding/geocoding.dart';
import 'package:get/get.dart';
import 'package:intl/intl.dart';
import 'package:tracelet/tracelet.dart' as tl;

import '../data/driver_tracking_location_modal.dart';
import '../data/start_end_trip_modal.dart';
import '../data/vehicle_tracking_request_model.dart';
import '../network/endpoints.dart';
import '../network/user_repo.dart';
import '../service/db_service.dart';
import '../utils/analytics_helper.dart';
import '../utils/common.dart';
import '../utils/driver_trackin_helper_utils.dart';
import '../utils/notification_helper.dart';
import '../utils/storage_helper.dart';



@pragma('vm:entry-point')
Future<void> traceletHeadlessTask(tl.HeadlessEvent event) async {
  WidgetsFlutterBinding.ensureInitialized();

  try {
    final name = event.name;
    final payload = event.event;

    log('[Tracelet HeadlessTask] name=$name event=$payload');

    switch (name) {
      case 'location':
        final coords = payload['coords'] as Map?;
        log(
          '[Tracelet HeadlessTask] location '
              'lat=${coords?['latitude']} lng=${coords?['longitude']} '
              'ts=${payload['timestamp']}',
        );
        break;

      case 'heartbeat':
        log('[Tracelet HeadlessTask] heartbeat event');
        break;

      case 'http':
[...]

PR #52, opened 2026-04-13
feat(dart,android,ios): speed-based motion detection mode

Sahilnenwani:
## Summary

Adds a second motion-detection strategy, selectable via
`MotionConfig.motionDetectionMode: speed`, that uses GPS speed from each
location fix to drive a **MOVING → SLOWING → STATIONARY** state machine
and atomically switch the native location engine between continuous
tracking and low-power periodic fixes (or geofences).

The default remains `accelerometer` — zero behavior change for existing
apps.

## Why

Accelerometer-based stop detection fails for vehicle tracking: a phone
resting on a smooth dashboard reads near-zero gravity-subtracted
magnitude at highway speed, so Tracelet falsely declares stationary.
The current workaround (`disableStopDetection: true`) keeps GPS on 24/7
and burns ~72%/day on always-on trackers (Life360 / fleet style).

A Dart-side state machine on `onLocation` works in foreground/background
but dies on app kill — the foreground service and background location
session survive, but the Dart VM doesn't. The state machine therefore
has to live natively. Expected battery use drops from ~72%/day to
~26%/day for a commute + overnight scenario.

### Public API

```dart
MotionConfig(
  motionDetectionMode: MotionDetectionMode.speed,    // default: accelerometer
  speedMovingThreshold: 1.5,                         // m/s, default 1.5 (≈5.4 km/h)
  speedStationaryDelay: 180,                         // s, red-light buffer
  stationaryTrackingMode: StationaryTrackingMode.periodic,
  stationaryPeriodicInterval: 120,                   // s between fixes when parked
  stationaryPeriodicAccuracy: DesiredAccuracy.high,
  speedWakeConfirmCount: 1,                          // instant wake at high accuracy
)

Tracelet.onMotionModeChange((SpeedMotionEvent e) {
  // e.previousState → e.state, e.trackingMode
});
```

### State machine

```
MOVING (continuous) ─speed<threshold─▶ SLOWING (continuous)
   ▲                                         │
   │                                  delay elapsed
   │                                         ▼
   └─wakeConfirmCount fixes≥threshold── STATIONARY (periodic | geofences)
```

State + counters + last-transition timestamp are persisted on every
change so process-death and reboot recovery resume the correct mode.

### Pigeon

- New enums `TlMotionDetectionMode { accelerometer, speed }` and
  `TlStationaryTrackingMode { periodic, geofences }`.
- New message `TlSpeedMotionEvent { state, previousState, trackingMode }`.
- New FlutterApi event `TraceletEventApi.onMotionModeChange(TlSpeedMotionEvent)`.

### Android

- New `SpeedMotionManager.kt` (peer of `MotionDetector`) — drives state
  transitions from `LocationEngine` fixes, persists to `StateManager`,
  emits events.
- `LocationService.kt` adds atomic `switchToStationaryPeriodic /
  switchToStationaryGeofences / switchToContinuous`. Periodic mode keeps
  the foreground service alive and fires one-shot `getCurrentLocation()`
  calls from a `Handler`-based timer — **no `SCHEDULE_EXACT_ALARM`
  permission, no WorkManager 15-min floor, doze-exempt**.
- `LocationEngine.kt` exposes `stopContinuous()` and
  `requestOneShotFix(accuracy, callback)` and forwards each fix's speed
  to the manager via `speedMotionSpeedSink`.
- `BootReceiver.kt`, `ConfigManager.kt`, `StateManager.kt`, `TraceletSdk.kt`,
  `TraceletEventSender.kt` (+ Listener / ListenerEventSender) updated.

### iOS

- New `SpeedMotionManager.swift` with `SpeedMotionDelegate` protocol —
  same state machine + persistence to `StateManager` UserDefaults.
  Negative `CLLocation.speed` (-1 invalid) is clamped to 0.
- `LocationEngine.swift` adds atomic switch methods that **preserve
  `CLServiceSession` / `CLBackgroundActivitySession`** across mode
  switches; only a user-initiated stop invalidates them.
- `TraceletSdk.swift` conditionally starts `SpeedMotionManager` instead
  of `MotionDetector` and restores persisted stationary state via
  `autoResumeTracking()`.
- `ConfigManager.swift`, `StateManager.swift`, `TraceletEventSending`,
  `DelegateEventSender`, `EventDispatcher.swift` updated.

### Backward compatibility

- Existing `onMotionChange(isMoving:)` still fires on MOVING ↔ STATIONARY
  transitions when speed mode is active.
- All new `MotionConfig` fields have defaults; existing configs are
  unchanged at runtime.
- `stopOnStationary: true` short-circuits to a full `stop()` instead of
  switching to stationary mode.

## Test plan

- [x] `melos run analyze` — clean across all 6 packages.
- [x] `flutter test` in `tracelet` — 460/460 pass (5 new MotionConfig
      speed-mode test cases + new `speed_motion_event_test.dart`).
- [x] `flutter test` in `tracelet_platform_interface` — 337/337 pass.
- [x] `flutter test` in `tracelet_android` and `tracelet_ios` — pass.
- [x] `melos run pigeon` — Pigeon bindings regenerated and committed.
- [ ] Android Gradle: `cd example/android && ./gradlew :tracelet_android:testDebugUnitTest`
      runs the new `SpeedMotionManagerTest.kt`.
- [ ] iOS: `cd sdk/ios && swift test` runs the new `SpeedMotionManagerTests.swift`.
- [ ] Manual device validation:
  1. Drive ~10 min → continuous fixes, `state: moving`.
  2. Park → after ~3 min, transition to `stationary`, fixes every 2 min.
  3. Drive away → first high-speed periodic fix wakes to continuous.
  4. Force-kill while parked → fixes continue (foreground service survives).
  5. Reboot while parked → `BootReceiver` resumes stationary periodic.

## Files

**Created**
- `packages/tracelet/lib/src/models/speed_motion_event.dart`
- `packages/tracelet/test/speed_motion_event_test.dart`
- `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/motion/SpeedMotionManager.kt`
- `sdk/android/tracelet-sdk/src/test/kotlin/com/ikolvi/tracelet/sdk/motion/SpeedMotionManagerTest.kt`
- `sdk/ios/Sources/TraceletSDK/motion/SpeedMotionManager.swift`
- `sdk/ios/Tests/TraceletSDKTests/SpeedMotionManagerTests.swift`
- `example/integration_test/speed_motion_test.dart`

**Modified**
- `packages/tracelet_platform_interface/pigeons/tracelet_api.dart` (+ regenerated `.g.{dart,kt,swift}`)
- `packages/tracelet_platform_interface/lib/src/{tracelet_platform,pigeon_tracelet,pigeon_event_receiver}.dart`
- `packages/tracelet/lib/{tracelet,src/tracelet,src/models/config}.dart`
- `packages/tracelet/test/models_test.dart`
- `packages/tracelet_android/.../flutter/EventDispatcher.kt`
- `packages/tracelet_ios/.../tracelet_ios/EventDispatcher.swift`
- `sdk/android/tracelet-sdk/.../{StateManager,ConfigManager,TraceletSdk,TraceletEventSender,TraceletListener,ListenerEventSender}.kt`
- `sdk/android/tracelet-sdk/.../location/LocationEngine.kt`
- `sdk/android/tracelet-sdk/.../service/LocationService.kt`
- `sdk/android/tracelet-sdk/.../receiver/BootReceiver.kt`
- `sdk/ios/Sources/TraceletSDK/{StateManager,ConfigManager,TraceletSdk,TraceletEventSending,DelegateEventSender}.swift`
- `sdk/ios/Sources/TraceletSDK/location/LocationEngine.swift`

## Notes for reviewers

- **Defaults are conservative on purpose.** `motionDetectionMode` defaults
  to `accelerometer`. Speed mode is opt-in.
- **Android scheduling choice.** Stationary periodic uses an in-process
  `Handler` timer on the existing foreground service rather than
  WorkManager (15-min floor) or AlarmManager exact alarms (permission
  + UX cost). Trade-off discussed in the plan.
- **`stationaryTrackingMode` is configurable** (`periodic` | `geofences`)
  rather than fixed, to support a future "monitor home geofence when
  parked" flow.


GalacticTitan (changes requested):
## Review — Speed-Based Motion Detection Mode

Thanks for this contribution, @Sahilnenwani! The feature itself is genuinely valuable — vehicle tracking false-stationarity is a real pain point, and the GPS-speed state machine approach is sound. The code is well-structured and follows the existing architecture properly. However, there are several issues that need addressing before this can be merged.

---

### Must-Fix (Medium Severity)

**1. Pigeon version bump (26.2.3 → 26.3.4) needs isolation**

The Pigeon upgrade regenerates ALL bindings (Swift `==`/`hash(into:)`, Kotlin codec byte values), inflating the diff by ~2000 lines. This makes it very hard to review the actual changes. Please either:
- Submit the Pigeon upgrade as a separate PR first, OR
- Confirm that the existing tests all pass with the regenerated bindings (Android Gradle + iOS Swift tests).

**2. Android `sendMotionChange` — verify map format**

In `SpeedMotionManager.kt`, the backward-compat `onMotionChange` fires:
```kotlin
events.sendMotionChange(mapOf("is_moving" to (newState != State.STATIONARY)))
```
Verify that `TraceletEventSender.sendMotionChange()` actually accepts a `Map<String, Any?>` with an `is_moving` key and constructs a proper `TlLocation` object. If it expects a location object directly, this will crash at runtime on MOVING↔STATIONARY transitions.

**3. iOS `emitSpeedMotionEvent` uses raw `[String: String]` dict**

The `SpeedMotionDelegate` protocol:
```swift
func emitSpeedMotionEvent(_ event: [String: String])
```
This bypasses the Pigeon-generated `TlSpeedMotionEvent` type. The wiring in `TraceletSdk.swift` has to manually construct the Pigeon message from a dictionary, which is fragile. Consider either:
- Passing a `TlSpeedMotionEvent` directly (the SDK module already imports the generated types), OR
- At minimum, adding a compile-time check that all required fields are present.

---

### Should-Fix (Low Severity)

**4. Android: Use `SystemClock.elapsedRealtime()` instead of `System.currentTimeMillis()`**

`System.currentTimeMillis()` is affected by NTP adjustments and manual clock changes. For the inter-fix interval EMA calculation in the SLOWING countdown, `SystemClock.elapsedRealtime()` is more robust.

**5. Input validation on speed config values**

`speedStationaryDelay = 0` and `speedWakeConfirmCount = 0` are accepted without bounds checking. While `0` delay is intentionally used in tests, production code should clamp to safe minimums (e.g., `speedWakeConfirmCount >= 1`, `speedStationaryDelay >= 0`). At minimum, add a `dart:developer` log warning for unusual values.

**6. Verify `stopOnStationary` interaction**

The PR description mentions `stopOnStationary: true` short-circuits to a full `stop()` instead of switching to stationary mode. Please confirm this code path exists in both `TraceletSdk.kt` and `TraceletSdk.swift`, and add a unit test for it.

---

### Incomplete Test Plan

Several items in your test plan are unchecked:
- [ ] Android Gradle unit tests (`SpeedMotionManagerTest.kt`)
- [ ] iOS Swift tests (`SpeedMotionManagerTests.swift`)
- [ ] All 5 manual device validation scenarios

Please run these and confirm results before re-requesting review.

---

### Summary

| Area | Status |
|------|--------|
| Architecture | ✅ Clean peer implementation |
| Backward compatibility | ✅ Default unchanged |
| Dart API and models | ✅ Well documented |
| Native state persistence | ✅ Process death + reboot recovery |
| Test coverage | ⚠️ Tests written but execution unverified |
| Pigeon upgrade side-effects | ⚠️ Needs isolation or verification |
| Runtime safety (issues 2, 3) | ❌ Potential crash paths |

Looking forward to the updates!

sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/motion/SpeedMotionManager.kt
```diff
@@ -0,0 +1,264 @@
...
+        state.speedLastTransition = System.currentTimeMillis()
+
+        // Update isMoving for compatibility with existing consumers
+        state.isMoving = newState != State.STATIONARY
+
+        // Emit speed motion change event
+        val eventData = mapOf(
+            "state" to newState.value,
+            "previousState" to previousState.value,
+            "trackingMode" to when (newState) {
```
GalacticTitan: **⚠️ Potential runtime crash**: `sendMotionChange()` in the existing codebase expects a `TlLocation`-shaped map (with `coords`, `battery`, `timestamp`, `is_moving`, `odometer`, etc.) to construct the Pigeon `TlLocation` message. Passing only `{"is_moving": true}` will likely cause a `NullPointerException` or `ClassCastException` when the EventSender tries to read missing fields.

**Suggestion**: Either pass the last known location through the speed motion callback chain, or add an overload like `sendMotionChangeMinimal(isMoving: Boolean)` that constructs a minimal location from the last cached fix + the new motion state.

sdk/ios/Sources/TraceletSDK/motion/SpeedMotionManager.swift
```diff
@@ -0,0 +1,256 @@
...
+/// Delegate that the `SpeedMotionManager` calls to switch the location engine
+/// between continuous and stationary modes and to emit Pigeon events.
+public protocol SpeedMotionDelegate: AnyObject {
+    func switchToContinuous()
+    func switchToStationaryPeriodic()
+    func switchToStationaryGeofences()
+    /// Emit a speed-motion state change event to Dart.
+    /// The generated Pigeon type `TlSpeedMotionEvent` is assumed to exist
+    /// (regenerate Pigeon after the schema update).
+    func emitSpeedMotionEvent(_ event: [String: String])
```
GalacticTitan: **Fragile type bridge**: This protocol method accepts a raw `[String: String]` dict, but the downstream Pigeon API expects a strongly-typed `TlSpeedMotionEvent`. Since the SDK module already imports the Pigeon-generated types, consider changing this to:

```swift
func emitSpeedMotionEvent(_ event: TlSpeedMotionEvent)
```

This way the compiler catches any missing/mistyped field at build time rather than at runtime. If you want to keep the SDK layer free of Pigeon types, at minimum document the required keys (`state`, `previousState`, `trackingMode`) and add a precondition/assert in the delegate implementation.

packages/tracelet_platform_interface/lib/src/generated/tracelet_api.g.dart
```diff
@@ -1,4 +1,4 @@
-// Autogenerated from Pigeon (v26.2.3), do not edit directly.
+// Autogenerated from Pigeon (v26.3.4), do not edit directly.
```
GalacticTitan: **Pigeon version upgrade side-effect**: This file changed from Pigeon `26.2.3` to `26.3.4`. The Pigeon upgrade regenerated all `==`, `hash(into:)`, and codec reader/writer methods across both iOS and Android generated files (~2000 lines of diff noise).

Please either:
1. Split the Pigeon upgrade into its own PR and confirm all existing tests pass, OR
2. Confirm the regenerated bindings pass all existing platform tests by running `./gradlew :tracelet_android:testDebugUnitTest` and `swift test` in the iOS SDK.

GalacticTitan:
@Sahilnenwani workflow is failing please fix it.

GalacticTitan (commented):
The workflow is failing can you check that?

Sahilnenwani:
Fixed the format issues

Sahilnenwani:
I also removed the pigeon upgrade code

GalacticTitan:
## Follow-up review

Thanks for the iteration — several earlier concerns are now resolved. Summary below.

### Addressed ✅
- Pigeon 26.2 → 26.3 upgrade reverted (diff size is now sane).
- Android `SpeedMotionManager` uses `SystemClock.elapsedRealtime()` for the EMA interval.
- Bounds clamping + warn logs added for `speedStationaryDelay` and `speedWakeConfirmCount`.
- iOS `SpeedMotionDelegate.emitSpeedMotionEvent` is now typed `(state:previousState:trackingMode:)` — no more raw `[String: String]`.
- `stopOnStationary` short-circuit implemented in both `TraceletSdk.kt` (lines 449 / 458) and `TraceletSdk.swift` (lines 1599 / 1630).

### Still blocking ❌

**1. Android event path is not type-safe (parity with iOS).**
Pigeon defines `TlSpeedMotionEvent`, but on Android:
- `TraceletEventSender.sendSpeedMotionChange(data: Map<String, Any?>)`
- `SpeedMotionManager` builds a stringly-typed `mapOf("state" to …, "previousState" to …, "trackingMode" to …)`

iOS was upgraded to typed parameters — Android should match. Right now a typo on any key silently sends a bad event. Please change the sender signature to take `TlSpeedMotionEvent` (or at least `state: String, previousState: String, trackingMode: String`) and construct the Pigeon message at the edge.

**2. Unexplained test deletions.** The diff deletes these files with no mention in the PR description:
- `example/ios/RunnerTests/PluginSecondaryEngineGuardTests.swift` (210 lines)
- `sdk/ios/Tests/TraceletSDKTests/LocationEngineChangePaceImmediateFixTests.swift` (133 lines)
- `sdk/ios/Tests/TraceletSDKTests/TraceletSdkTests.swift` (~50 lines removed)

Please either justify the removals in the PR description or restore them. If they became obsolete because of the new mode-switch atomicity, call that out and replace with equivalent coverage on the new paths.

**3. Native CI not green yet.** `Build Android`, `Build iOS`, and `Coverage Report` are still `pending`. The new `SpeedMotionManagerTest.kt` and `SpeedMotionManagerTests.swift` are therefore unverified in CI. Please re-request review after those turn green.

**4. Missing CHANGELOG entries.** This is a new public API (`MotionConfig.motionDetectionMode`, `SpeedMotionEvent`, `Tracelet.onMotionModeChange`, new Pigeon enums). Add `**FEAT**:` entries to:
- `packages/tracelet/CHANGELOG.md`
- `packages/tracelet_platform_interface/CHANGELOG.md`
- `packages/tracelet_android/CHANGELOG.md`
- `packages/tracelet_ios/CHANGELOG.md`

### Nice-to-have
- Consider a small `SpeedFixListener` interface on `LocationEngine` instead of the free `speedMotionSpeedSink` property — makes the coupling explicit and easier to mock.
- PR description's manual device validation checklist is still all-unchecked. Even a brief note confirming you ran scenarios 1–3 in an emulator with mocked speed would be valuable.

### Backward compatibility — OK
Defaults preserve existing behaviour, new Pigeon fields/enums are additive, `onMotionChange(isMoving:)` still fires under speed mode. Federated publish order + version pin in `tracelet` should keep existing consumers unbroken.

Architecture itself is sound — state machine, persistence, and atomic engine switching all look right. It's mainly the typed-event parity, test deletions, and green CI that I need before approving.

GalacticTitan:
Hi @Sahilnenwani — I did a deeper file-by-file audit of this PR and found a major problem. **This branch is based on `v1.8.10` (commit `5c594a7`)**, before #51, #54, and #55 ever landed on `main`. As a result, the diff GitHub shows against current `main` contains mass-reversions of fixes from three subsequent releases.

**This PR cannot be merged as-is — it would silently undo three releases of work.**

### What the current diff would revert

| Lost on merge | Origin commit / PR |
|---|---|
| Secondary FlutterEngine guard (Android + iOS) | `397b6bd` (#51) — deletes `PluginSecondaryEngineGuardTests.swift` (210L), `PluginSecondaryEngineGuardTest.kt`, `geofence_start_mode_test.dart` |
| Geofence + location `extras` JSON round-trip | `a9cb5bf` (#51 follow-up) — `TraceletDatabase.kt` reverts to broken `Map.toString()`; deletes `GeofenceExtrasRoundTripTest.kt` (89L) and `LocationMapRoundTripTest.kt` (30L) |
| First-fix latency perf (changePace one-shot fix) | `cdd9227` (#54) — deletes `LocationEngineChangePaceImmediateFixTest.kt` (154L) and `LocationEngineChangePaceImmediateFixTests.swift` (133L); `MotionDetector.kt` -35L removes the post-`changePace` re-engagement; `LocationEngine` perf one-shot removed |
| CHANGELOG entries for v1.8.11, v1.8.12, v1.8.13 | Erased from `tracelet`, `tracelet_android`, `tracelet_ios` |
| Version numbers | `sdk/android/gradle.properties`: `SDK_VERSION=1.0.12 → 1.0.10`; `tracelet_android/android/build.gradle`: `tracelet-sdk:1.0.12 → 1.0.10`; `sdk/ios/TraceletSDK.podspec`: `1.0.11 → 1.0.7` (off by 4!) |

### What you need to do

```bash
git fetch origin
git checkout feat/speed-motion
git rebase origin/main
```

Since the speed-motion changes touch `LocationEngine`, `MotionDetector`, `TraceletSdk`, and `LocationService` — exactly the files modified by #54 — there will be **real merge conflicts** that need to be resolved by combining both feature sets, not picking one.

After rebase, the diff should contain only:
- The 7 new speed-motion files (`SpeedMotionManager.{kt,swift}` + tests + `speed_motion_event.dart` + integration test).
- Additive Pigeon schema (`TlMotionDetectionMode`, `TlStationaryTrackingMode`, `TlSpeedMotionEvent`, `onMotionModeChange`) and regenerated bindings.
- Modifications to `LocationEngine`, `LocationService`, `MotionDetector`, `TraceletSdk`, `ConfigManager`, `StateManager`, `BootReceiver` that **add** speed-mode hooks **without** removing the existing perf/extras/guard code.
- New CHANGELOG `**FEAT**:` entries (additive) in all 4 affected packages.
- **No version downgrades**, no deleted CHANGELOG sections, no deleted `extras` JSON code, no deleted secondary-engine-guard tests.

The earlier feedback (typed `TlSpeedMotionEvent` on Android, native CI green, etc.) still stands and can be addressed in the same rebase.

Force-push the rebased branch and re-request review. Happy to help verify the post-rebase diff if useful.

GalacticTitan (changes requested):
Requesting changes — see [my detailed comment above](https://github.com/Ikolvi/Tracelet/pull/52#issuecomment-4276182122) for the full audit. **Do not merge** until this branch has been rebased onto current `main`.

Blocking issues:

1. Branch is based on `v1.8.10` (5c594a7); diff against current main mass-reverts:
   - All of #51 (secondary FlutterEngine guard + geofence/location `extras` JSON round-trip)
   - All of #54 (changePace one-shot perf fix + post-changePace MotionDetector re-engagement)
   - CHANGELOG entries for v1.8.11, v1.8.12, v1.8.13 across 3 packages
   - Native SDK versions: Android `SDK_VERSION=1.0.12 → 1.0.10`, iOS `1.0.11 → 1.0.7`
2. Android `sendSpeedMotionChange(Map<String, Any?>)` should use the typed `TlSpeedMotionEvent` Pigeon message for parity with iOS.
3. Native CI (`Build Android`, `Build iOS`) was still pending at last check, so the new `SpeedMotionManagerTest{,s}` haven't been run in CI.
4. No CHANGELOG entries added for the new public API.

Please rebase onto `origin/main`, resolve the real conflicts in `LocationEngine` / `MotionDetector` / `TraceletSdk` / `LocationService` by **combining** speed-motion with the existing perf and extras code (not picking one), then force-push and re-request review.

GalacticTitan:
# Full Audit Report — PR #52 (`feat/speed-motion`)

## Verdict: ❌ Changes requested. Do not merge.

The branch is **based on tag `v1.8.10` (commit `5c594a7`, 2026-04-13)** and has never been rebased. Three subsequent PRs landed on `main` since then, so the diff GitHub shows against current `main` contains mass-reversions of merged work in addition to the legitimate speed-motion feature.

```
Branch base:  5c594a7  (v1.8.10, 2026-04-13)
Current main: 58af70d  (revert of #55)
Commits on main since branch base: 11 (4 substantive, 5 benchmark CI, 1 bump+revert)
```

Until the branch is rebased onto current `main`, no further code review can be reliably done because every modified file conflates two unrelated change sets.

---

## 1. Mass-reversion of merged work (BLOCKING)

These deletions/regressions are not part of the speed-motion feature — they are the result of a stale base.

### 1.1 #51 — Secondary FlutterEngine guard (`397b6bd`)
| File | Status in PR diff | Lines lost |
|---|---|---|
| `example/ios/RunnerTests/PluginSecondaryEngineGuardTests.swift` | DELETED | 210 |
| `packages/tracelet_android/android/src/test/kotlin/com/ikolvi/tracelet/flutter/PluginSecondaryEngineGuardTest.kt` | DELETED | (entire file) |
| `example/integration_test/geofence_start_mode_test.dart` | DELETED | (entire file) |
| Production guard code in plugin singletons | REVERTED | — |

This guard prevents Firebase background messaging (and any other secondary `FlutterEngine`) from overwriting the SDK's event sink and silently killing geofence callbacks. Reverting it brings back a real production bug.

### 1.2 #51 follow-up — Geofence + location `extras` JSON round-trip (`a9cb5bf`)
`sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/db/TraceletDatabase.kt` reverts from JSON serialization to `Map.toString()`:

```kotlin
// Current main (correct):
val extrasJson: String? = if (extrasMap != null) {
    org.json.JSONObject(extrasMap.mapKeys { it.key.toString() }).toString()
} else null
put(COL_EXTRAS, extrasJson)

// PR diff (regression):
put(COL_EXTRAS, location["extras"]?.toString())   // produces "{key=value}"
                                                  // unparseable on read-back
```

Plus deletion of:
| File | Lines lost |
|---|---|
| `sdk/android/tracelet-sdk/src/test/kotlin/com/ikolvi/tracelet/sdk/db/GeofenceExtrasRoundTripTest.kt` | 89 |
| `sdk/android/tracelet-sdk/src/test/kotlin/com/ikolvi/tracelet/sdk/db/LocationMapRoundTripTest.kt` | 30 |

Net effect: `GeofenceEvent.extras` becomes empty in production, identical to the bug fixed in #51.

### 1.3 #54 — First-fix latency perf (`cdd9227`)
| File | Change in PR diff | Effect |
|---|---|---|
| `sdk/android/.../LocationEngine.kt` | Removes one-shot `getCurrentLocation()` on `changePace(true)` | First fix latency goes back to 11–50 s |
| `sdk/ios/.../location/LocationEngine.swift` | Removes one-shot `requestLocation()` on `changePace(true)` | Same regression on iOS |
| `sdk/android/.../motion/MotionDetector.kt` | -35 lines | Removes accelerometer + significant-motion re-engagement after manual `changePace(false)` — Android SDK can get stuck permanently stationary |
| `sdk/android/tracelet-sdk/src/test/kotlin/com/ikolvi/tracelet/sdk/location/LocationEngineChangePaceImmediateFixTest.kt` | DELETED 154 lines | Lost coverage |
| `sdk/ios/Tests/TraceletSDKTests/LocationEngineChangePaceImmediateFixTests.swift` | DELETED 133 lines | Lost coverage |

### 1.4 CHANGELOG erasure
Entries for **v1.8.11**, **v1.8.12**, **v1.8.13** are removed from:
- `packages/tracelet/CHANGELOG.md`
- `packages/tracelet_android/CHANGELOG.md`
- `packages/tracelet_ios/CHANGELOG.md`
- `sdk/ios/CHANGELOG.md` (1.0.11 entry)

### 1.5 Native SDK version downgrades
| File | Change | Severity |
|---|---|---|
| `sdk/android/gradle.properties` | `SDK_VERSION=1.0.12 → 1.0.10` | Downgrade |
| `packages/tracelet_android/android/build.gradle` | `tracelet-sdk:1.0.12 → 1.0.10` | Downgrade |
| `sdk/ios/TraceletSDK.podspec` | `1.0.11 → 1.0.7` | **Off by 4 — would break consumers** |

---

## 2. Code-quality issues in the actual feature changes (FIXABLE)

These remain valid concerns even after rebase.

### 2.1 Type-safety regression on Android event path (Golden Rule #2 violation)
Pigeon defines `TlSpeedMotionEvent` with three typed fields. iOS was wired up correctly:
```swift
public func emitSpeedMotionEvent(state: String, previousState: String, trackingMode: String)
```

But Android still uses a stringly-typed map:
```kotlin
// TraceletEventSender.kt:48
fun sendSpeedMotionChange(data: Map<String, Any?>)

// SpeedMotionManager.kt:253-265
val eventData = mapOf(
    "state" to newState.value,
    "previousState" to previousState.value,
    "trackingMode" to when (newState) { ... },
)
events.sendSpeedMotionChange(eventData)
```

A typo on any key silently sends a malformed event. Use the Pigeon-typed message (or at minimum typed parameters like iOS) and construct the Pigeon `TlSpeedMotionEvent` at the dispatcher edge.

### 2.2 Native unit tests unverified in CI
`Build Android` and `Build iOS` jobs were still pending at last check. The new `SpeedMotionManagerTest.kt` (272 L) and `SpeedMotionManagerTests.swift` (192 L) have therefore not been validated by CI yet. Author's own checklist still has these unchecked:
> [ ] Android Gradle: ./gradlew :tracelet_android:testDebugUnitTest
> [ ] iOS: cd sdk/ios && swift test
> [ ] All 5 manual device validation scenarios

### 2.3 Missing CHANGELOG entries for the new public API
After rebase, the PR adds these new public surfaces:
- `MotionConfig.motionDetectionMode` + 5 related fields
- `MotionDetectionMode`, `StationaryTrackingMode`, `SpeedMotionState`, `SpeedMotionTrackingMode` enums
- `SpeedMotionEvent` model
- `Tracelet.onMotionModeChange((SpeedMotionEvent e) {...})`
- Pigeon: `TlMotionDetectionMode`, `TlStationaryTrackingMode`, `TlSpeedMotionEvent`, `TraceletEventApi.onMotionModeChange`

Add `**FEAT**:` entries to:
- `packages/tracelet/CHANGELOG.md`
- `packages/tracelet_platform_interface/CHANGELOG.md`
- `packages/tracelet_android/CHANGELOG.md`
- `packages/tracelet_ios/CHANGELOG.md`

### 2.4 Coupling — `speedMotionSpeedSink`
`LocationEngine` exposes a public mutable closure property that `SpeedMotionManager` writes to. Acceptable, but a `SpeedFixListener` interface would make the contract explicit, easier to unit-test, and prevent accidental clobbering.

---

## 3. What's actually good ✅

- Architecture is correct: native state machine peer to `MotionDetector`, atomic engine switch, `StateManager` persistence for process-death + `BootReceiver` recovery — this is the right design.
- Defaults preserve existing behaviour (`motionDetectionMode = accelerometer`).
- Backward-compat `onMotionChange(isMoving:)` still fires under speed mode.
- Pigeon additions are appended (not reordered) — wire-format compatible.
- Bounds clamping with warn logs for `speedStationaryDelay` / `speedWakeConfirmCount`.
- Android uses `SystemClock.elapsedRealtime()` (immune to NTP) for the EMA interval.
- iOS clamps `CLLocation.speed = -1` (invalid) to 0.
- `stopOnStationary` short-circuit implemented on both platforms (`TraceletSdk.kt#L449/458`, `TraceletSdk.swift#L1599/1630`).
- iOS `SpeedMotionDelegate.emitSpeedMotionEvent` already uses typed parameters (issue #2.1 only affects Android).

---

## 4. Required actions before re-review

1. **Rebase**:
   ```bash
   git fetch origin
   git checkout feat/speed-motion
   git rebase origin/main
   ```
   Real conflicts will surface in `LocationEngine.{kt,swift}`, `MotionDetector.kt`, `TraceletSdk.{kt,swift}`, `LocationService.kt`. Resolve them by **combining** speed-motion with existing perf/extras/guard code — never picking one over the other. Drop *all* CHANGELOG and version downgrades from the branch.

2. **Fix Android typed event** (issue 2.1): change the Android sender signature to take `TlSpeedMotionEvent` directly (parity with iOS).

3. **Add CHANGELOG entries** in the 4 affected packages with `**FEAT**:` prefix for the new public API.

4. **Wait for `Build Android` + `Build iOS` to be green** so the new native unit tests are actually validated, and tick the test plan checkboxes in the PR description.

5. **Force-push** and re-request review.

---

## 5. Post-rebase expected diff shape (sanity check before re-requesting review)

After rebase, `git diff origin/main..HEAD --stat` should show approximately:

- 7 added files: `SpeedMotionManager.{kt,swift}` + 2 native tests + `speed_motion_event.dart` + `speed_motion_event_test.dart` + `example/integration_test/speed_motion_test.dart`
- ~12 modified files: Pigeon schema + 3 regenerated bindings + `MotionConfig` model + `tracelet.dart` + `EventDispatcher.{kt,swift}` + native `ConfigManager`, `StateManager`, `TraceletSdk`, `LocationEngine`, `LocationService`, `BootReceiver` (additive hooks only)
- **0 deleted files**
- **0 deleted CHANGELOG sections**
- **0 lowered version numbers**

If your post-rebase diff still shows deleted files or lowered versions, the rebase resolved a conflict the wrong way and needs to be redone.

---

Happy to verify the post-rebase diff before you re-request review. The feature itself is wanted and the architecture is sound — this just needs a clean base.

GalacticTitan:
## Code Review — v2.0.0 Post-Rebase Assessment

Thanks for the contribution @Sahilnenwani — the speed-based motion detection feature is well-designed. Here is a detailed review against the current `main` (2.0.0).

---

### ✅ What looks good

- **Architecture is sound**: State machine (`MOVING → SLOWING → STATIONARY → MOVING`) correctly implemented on both Android and iOS with process-death persistence via `StateManager`.
- **Android uses `SystemClock.elapsedRealtime()`** (immune to NTP changes) for the EMA interval — correct choice.
- **iOS clamps `CLLocation.speed = -1`** (invalid) to 0 — handles edge case properly.
- **Typed Android event fixed**: `sendSpeedMotionChange(state: String, previousState: String, trackingMode: String)` — the untyped `Map` concern from the previous review has been addressed. 👍
- **Backward compatible**: Default `motionDetectionMode = accelerometer` means existing apps are unaffected. `onMotionChange(isMoving:)` still fires under speed mode.
- **Equality/hashCode**: `MotionConfig` correctly includes all 7 new fields.
- **Test coverage is excellent**: Both platforms have recording-double-based unit tests covering state transitions, wake confirmation, negative speed, geofence mode, and persistence. Dart model round-trip tests included.
- **Bounds clamping with warn logs** for `speedStationaryDelay` / `speedWakeConfirmCount`.

---

### 🔴 Must fix before merge

**1. Rebase onto `main`** — The PR is currently `CONFLICTING`. Tracelet 2.0.0 landed a major config restructuring (`GeoConfig`, `AndroidConfig`, `IosConfig` split) and significant changes to `TraceletSdk.kt`, `TraceletSdk.swift`, `LocationEngine`, and the Pigeon schema. Expected conflict zones:

- `packages/tracelet/lib/src/models/config.dart` (MotionConfig restructured)
- `sdk/android/.../TraceletSdk.kt` / `sdk/ios/.../TraceletSdk.swift` (lifecycle changes)
- `sdk/android/.../LocationEngine.kt` / `sdk/ios/.../LocationEngine.swift`
- `packages/tracelet_platform_interface/pigeons/tracelet_api.dart` (Pigeon schema)

After resolving conflicts, **regenerate Pigeon bindings** (`dart run pigeon`).

**2. CHANGELOG entries** — Need to be written for the 2.0.0 context across all 4 packages (`tracelet`, `tracelet_platform_interface`, `tracelet_android`, `tracelet_ios`).

---

### ⚠️ Should fix

**3. iOS/Android API asymmetry** — Android `TraceletEventSender` uses typed parameters:
```kotlin
fun sendSpeedMotionChange(state: String, previousState: String, trackingMode: String)
```
But iOS `TraceletEventSending` uses an untyped dictionary:
```swift
func sendSpeedMotionEvent(_ data: [String: Any])
```
Recommend making iOS typed to match Android for consistency and type safety.

**4. iOS `DelegateEventSender.sendSpeedMotionEvent` is a no-op** — The standalone SDK delegate path silently drops speed motion events. Either add a `tracelet(_:didChangeSpeedMotion:)` method to `TraceletDelegate`, or document that speed motion events are only available via the Flutter plugin.

---

### 🟢 Nice to have (non-blocking)

**5. `speedMotionSpeedSink` coupling** — `LocationEngine` exposes a public mutable closure that `SpeedMotionManager` writes to. A `SpeedFixListener` interface/protocol would make the contract explicit and prevent accidental clobbering.

**6. Dart-side config validation** — Speed config fields log warnings for out-of-range values on the native side, but the Dart `MotionConfig` constructor doesn't validate. Consider adding `assert` or `clamp` for `speedMovingThreshold`, `speedStationaryDelay`, `speedWakeConfirmCount`.

---

### Summary

| Aspect | Rating |
|---|---|
| Architecture | ⭐⭐⭐⭐⭐ |
| Code quality | ⭐⭐⭐⭐ |
| Test coverage | ⭐⭐⭐⭐⭐ |
| Platform parity | ⭐⭐⭐⭐ (iOS typing gap) |
| Merge readiness | 🔴 Needs rebase onto 2.0.0 |

The feature is wanted and the architecture is solid. Main blocker is the rebase — 2.0.0 changed a lot of the files this PR touches. After rebase + Pigeon regen + CHANGELOG update, this is ready to merge. Happy to help verify the post-rebase diff.

GalacticTitan (changes requested):
Please rebase to 2.0.0 and fix rhe mentioned changes

PR #55, opened 2026-04-19
chore: pigeon version update

Sahilnenwani:
## Summary

- Upgrade Pigeon code generator from v26.2.3 to v26.3.4
- Regenerate all platform bindings (Dart, Kotlin, Swift) with the new version
- Update `pigeon` constraint in `tracelet_platform_interface/pubspec.yaml`

## Motivation

Isolates the Pigeon version bump into its own PR so the regenerated
codec, equality, and hash changes do not inflate the diff of feature PRs.
This was requested during review of the speed-motion feature branch.

## Changed files

| File | Change |
|---|---|
| `tracelet_platform_interface/pubspec.yaml` | `pigeon: ^26.1.7` to `^26.3.4` |
| `tracelet_api.g.dart` | Regenerated with Pigeon v26.3.4 |
| `TraceletApi.g.kt` | Regenerated with Pigeon v26.3.4 |
| `TraceletApi.g.swift` | Regenerated with Pigeon v26.3.4 |

## How it was generated

Per [CONTRIBUTING.md](./CONTRIBUTING.md):

```bash
# 1. Updated pigeon constraint in pubspec.yaml
# 2. Resolved dependencies
flutter pub get
# 3. Regenerated bindings
melos run pigeon
# 4. Formatted
melos run format:fix
```

## Test plan

- [x] `melos run format` passes (0 files changed)
- [x] `melos run analyze` passes (no issues across 6 packages)
- [x] `flutter test --coverage` passes (820 tests across all packages)
- [ ] CI `build-android` passes
- [ ] CI `build-ios` passes


GalacticTitan:
Heads-up — this merge had to be reverted via #56.

### What happened

The Pigeon regen overwrote a documented manual fix in `packages/tracelet_platform_interface/lib/src/generated/tracelet_api.g.dart` for `getPrivacyZones()`:

```dart
// MANUAL FIX: Pigeon generates .cast<Map<String, Object?>>() which is a
// lazy cast that fails at runtime because platform channel maps arrive as
// Map<Object?, Object?> and Dart's reified generics reject the cast.
// Eagerly deep-cast each map instead.  Re-apply after Pigeon regeneration.
return (pigeonVar_replyValue! as List<Object?>)
    .map((e) => Map<String, Object?>.from(e! as Map))
    .toList();
```

After the regen, this reverted to the broken `.cast<Map<String, Object?>>()` form, which throws at runtime in production:
```
type '_Map<Object?, Object?>' is not a subtype of type 'Map<String, Object?>'
```

The fix is documented in `packages/tracelet_platform_interface/CHANGELOG.md`:
> **FIX**: Fix Pigeon-generated `getPrivacyZones()` lazy `.cast<Map<String, Object?>>()` — eagerly deep-cast each map to prevent `_Map<Object?, Object?>` type error at runtime.

### Why CI didn't catch it

`pigeon_tracelet_test.dart` uses a mock that returns `Map<String, Object?>` directly, bypassing the actual platform-channel deserialization where the `_Map<Object?, Object?>` runtime type appears. The bug only manifests on a real device.

### Audit of the rest of the diff

I checked all three regenerated files — Dart, Kotlin, Swift — for any other `MANUAL FIX` / workaround / hack markers and any other lazy `.cast<Map<...>>()` patterns. The `getPrivacyZones()` overwrite was the **only** regression. The Pigeon 26.3.4 codec changes themselves (improved `-0.0`/`NaN` equality + hash, etc.) are good.

### Path forward for re-doing this PR

1. Bump Pigeon to 26.3.4 + regenerate.
2. **Re-apply the manual fix** at `getPrivacyZones()` in the regenerated `tracelet_api.g.dart` (preserve the comment block so the next regen reviewer sees it).
3. Add a real platform-channel round-trip test for `getPrivacyZones()` that exercises the actual codec path so this regression can't return.
4. Add a `**CHORE**: bump pigeon to 26.3.4` entry to `packages/tracelet_platform_interface/CHANGELOG.md`.

This was my mistake for merging without re-checking the privacy-zone area against the documented CHANGELOG fix. Apologies @Sahilnenwani — the Pigeon bump itself is still wanted, just needs the manual fix re-applied.

PR #64, opened 2026-04-30
Fix bugs - Add enhancements

Moo-Meshrif:
🚀 Technical Summary
1. Android: Smart Notification System
Commit: feat(android): implement smart notification system with fast foreground detection and hardened restoration logic

Atomic Suppression: Re-engineered onStartCommand to perform an "atomic" promotion/demotion (startForeground + stopForeground). This satisfies Android 14+ requirements without showing the notification when the app is in the foreground.
Dual-Layer Detection: Combines ProcessLifecycleOwner with ActivityManager.getMyMemoryState() for instant, reliable process importance checking.
Smart Restoration: Added state tracking (isForegroundService) to automatically restore the notification if it was manually dismissed when the app moves to the background.
Anti-Flicker Protection: Prevents redundant notify() calls, eliminating notification flashes and unnecessary sound/vibration alerts during lifecycle transitions.
Local SDK Linking: Implemented a conditional includeBuild in Gradle to automatically prioritize local SDK source code during development.

2. iOS: Motion Sensor Initialization Fix
Commit: fix(ios): prevent lazy initialization of motion sensors in stop() when disabled

Lazy Guard: Introduced isFullModeStarted flag to prevent MotionDetector from initializing sensors during the stop() sequence.
Resource Management: Prevents unnecessary memory allocation and potential crashes when stopping a service that was partially started or disabled via config.
Permission Safety: Ensures that stopping the service doesn't trigger unexpected permission prompts on iOS.

### DIFF
--- a/packages/tracelet/lib/src/models/config.dart
+++ b/packages/tracelet/lib/src/models/config.dart
@@ -1,4 +1,4 @@
-import 'package:meta/meta.dart';
+import 'package:flutter/foundation.dart';
 import 'package:tracelet_platform_interface/tracelet_platform_interface.dart';
 
 import '_helpers.dart';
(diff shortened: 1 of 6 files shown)


GalacticTitan:
Merged manually to main

PR #266, opened 2026-07-27
feat(android): add requestTermination for headless isolate GPS shutdown

### LINKED ISSUE #267: [Feature]: Expose requestTermination on headless MethodChannel
### Is your feature request related to a problem?

When an FCM silent push triggers a background task (e.g. auto-punchout) while the app is terminated, the background handler cannot stop `TraceletEngine` because `Tracelet.stop()` uses Pigeon which is unavailable from headless Dart isolates. The GPS foreground service keeps polling and draining battery until the user manually opens the app.

### Describe the solution you'd like

Add a `requestTermination` handler to the `com.tracelet/methods` MethodChannel in `HeadlessTaskService.kt` that calls `TraceletSdk.stop()`. Expose it as `Tracelet.requestTermination()` — a static method that uses the headless MethodChannel directly, bypassing Pigeon.

### Describe alternatives you've considered

Setting a `SharedPreferences` flag and waiting for the foreground app to call `forceStop()` on next resume. This works but wastes battery in the interim — GPS keeps polling with no upload.

GhagSagar23:
## Summary

- Add `"requestTermination"` handler to `HeadlessTaskService.kt` methods channel, calling `TraceletSdk.getInstance(context).stop()` with try/catch error handling
- Add `Tracelet.requestTermination()` static method that invokes the headless MethodChannel directly, with `MissingPluginException` catch for iOS graceful fallback
- Add unit tests for the new method (success, iOS fallback, null response)
- Android only — iOS headless architecture differs and is out of scope

Closes #267

## Why

When an FCM silent push triggers a background task (e.g. auto-punchout) while the app is killed, the background handler cannot stop `TraceletEngine` because `Tracelet.stop()` uses Pigeon which doesn't work from headless isolates. The GPS foreground service keeps running, draining battery until the user manually opens the app.

`requestTermination()` uses the existing `com.tracelet/methods` MethodChannel (same one used by `setDynamicHeaders` and `setSyncBodyResponse`) which IS available in headless isolates, routing directly to the native `TraceletSdk.stop()`.

## Usage

```dart
// In FCM background handler or other headless isolate:
final stopped = await Tracelet.requestTermination();
// Returns true if native service acknowledged, false on iOS or failure
```

## Test plan

- [x] Unit tests for `Tracelet.requestTermination()` (3 cases: success, MissingPluginException, null)
- [ ] Verify Kotlin compiles via CI Gradle build
- [ ] `melos run analyze` clean on `packages/tracelet`
- [ ] `melos run test` passes
- [ ] Manual test: trigger FCM silent push while app killed, verify GPS service stops
- [ ] Verify iOS gracefully returns `false` (no crash)

### DIFF
--- a/packages/tracelet/lib/src/tracelet.dart
+++ b/packages/tracelet/lib/src/tracelet.dart
@@ -271,6 +271,24 @@ class Tracelet {
     return _stateFromMap(result);
   }
 
+  /// Request the native service to terminate from a headless isolate.
+  ///
+  /// Unlike [stop], this uses the headless MethodChannel directly,
+  /// so it works from FCM background handlers where Pigeon is unavailable.
+  /// Returns whether the native service acknowledged the request.
+  /// No [State] is returned since Pigeon is unavailable in headless isolates.
+  ///
+  /// Android only. On iOS, returns `false` (headless architecture differs).
+  static Future<bool> requestTermination() async {
+    try {
+      const channel = MethodChannel('com.tracelet/methods');
+      final result = await channel.invokeMethod<bool>('requestTermination');
+      return result ?? false;
+    } on MissingPluginException {
+      return false;
+    }
+  }
+
   /// Start geofence-only tracking mode.
   ///
   /// The plugin will only monitor geofences without continuous location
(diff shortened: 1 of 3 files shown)


GalacticTitan (approved):
Reviewed: channel name matches METHODS_CHANNEL_NAME, TraceletSdk.stop() exists, imports present, mirrors existing setDynamicHeaders/setSyncBodyResponse handler pattern. LGTM.

Issue #264, opened 2026-07-26
[Bug]: Android startOnBoot geofence restart crashes with uninitialized geofenceManager


ezechukwu69:
### Package Name

tracelet (Core Plugin / Android SDK)

### Tracelet Version

3.6.12 (`tracelet`, `tracelet_android`, and `com.ikolvi:tracelet-sdk`)

### Platform

Android

### Device & OS Version

Reproduced on an Android emulator running Android 17 / API 37. The failure is in the native boot path before Flutter/Dart application startup.

### Description

With geofence tracking persisted as enabled, `startOnBoot: true`, and `stopOnTerminate: false`, the app process can crash while Android starts `LocationService` after a reboot/process restart.

The SDK starts heavy initialization asynchronously, but `bootstrapForBackground()` does not wait for it before `startBootTracking()` accesses `TraceletSdk.geofenceManager`. App-side serialization around `Tracelet.ready()`, `getState()`, and `startGeofences()` cannot prevent the crash because it happens before Dart runs.

### Steps to Reproduce

1. Configure Tracelet:

```dart
await Tracelet.ready(
  Config.balanced().copyWith(
    app: const AppConfig(
      stopOnTerminate: false,
      startOnBoot: true,
    ),
  ),
);
```

2. Grant background location permission.
3. Call `Tracelet.startGeofences()`.
4. Confirm native state has `enabled=true`, `trackingMode=1`.
5. Reboot the device/emulator.
6. Observe `LocationService` crash during `startBootTracking()`.

Persisted state from the reproduction:

```xml
<boolean name="didLaunchInBackground" value="true" />
<int name="trackingMode" value="1" />
<boolean name="didDeviceReboot" value="true" />
<boolean name="enabled" value="true" />
```

### Stack Trace

```text
java.lang.RuntimeException: Unable to start service
  com.ikolvi.tracelet.sdk.service.LocationService

Caused by: kotlin.UninitializedPropertyAccessException:
  lateinit property geofenceManager has not been initialized

  at com.ikolvi.tracelet.sdk.TraceletSdk.getGeofenceManager(TraceletSdk.kt:101)
  at com.ikolvi.tracelet.sdk.service.LocationService.startBootTracking(LocationService.kt:1152)
  at com.ikolvi.tracelet.sdk.service.LocationService.onStartCommand(LocationService.kt:630)
```

### Root Cause

[...]

PR #270, opened 2026-07-28
fix(android): await init on every boot/broadcast entry point (geofenceManager lateinit crash, regression from #260)

dagovalsusa:
## Problem

Regression from #260 (moving `initialize()` off the main thread). Since that change, `TraceletSdk.initialize()` runs its heavy setup — opening the Rust `DatabaseManager` and wiring the `lateinit` managers (`geofenceManager`, `locationEngine`, …) — on a background `tracelet-init` thread and returns immediately. Entry points that were synchronous before, and that dereference those managers right after calling `initialize()`, now race that thread.

Observed in production (Android, fatal):

```
Fatal Exception: lateinit property geofenceManager has not been initialized
  at TraceletSdk.getGeofenceManager
  at LocationService.startBootTracking
  at LocationService.onStartCommand
```

…with the `tracelet-init` thread still inside `initializeInternal()` at crash time.

`ready()` already waited on the init latch, but the **native boot / broadcast paths do not go through `ready()`**, so they were unguarded.

## Affected call sites

- **`LocationService.startBootTracking`** → `bootstrapForBackground()` was gated only on `rustDatabase == null`. `rustDatabase` is assigned *early* in `initializeInternal`, well before `geofenceManager`, so the guard could pass while managers were still unset; `startBootTracking` then reads `sdk.geofenceManager` → crash.
- **`GeofenceBroadcastReceiver.onReceive`** — `initialize()` then immediately reads `geofenceManager`. Now throws, is caught, `manager` becomes `null`, and the geofence transition is **silently dropped** — a cold-boot enter/exit (a trip start) is lost.
- **`CrashConfirmReceiver`** — `initialize()` then `deliverConfirmedImpact()` on not-yet-wired state.

`PeriodicLocationWorker` is unaffected (routes through `bootstrapForBackground`, and its direct `locationEngine` read is already `isReady`-gated).

## Fix

Centered on a single gate:

- `awaitInit()` now returns `Boolean` — `true` when subsystems are safe to touch, `false` on timeout or init failure. Timeout raised 15s → 30s.
- `initializeInternal()` failure sets an `initFailed` flag and logs the **full stacktrace** (was only `t.message`), so `awaitInit()` reports failure instead of letting callers hit a later, misleading `UninitializedPropertyAccessException`.
- Every synchronous entry point routes through `initialize()` + `awaitInit()` and **bails** (skip / return) when it returns `false`, rather than dereferencing: `ready()`, `bootstrapForBackground()`, `startBootTracking`'s geofence wiring, `GeofenceBroadcastReceiver`, `CrashConfirmReceiver`.
- `ready()` now also calls `initialize()` first (previously relied on `onAttachedToEngine` having called it).

No self-deadlock: `awaitInit()` only runs on service / broadcast / plugin threads, never on `tracelet-init`.

## Known limitation (pre-existing, not addressed here)

`ready()` runs on the platform main thread and blocks on `awaitInit()`; #260 moved the DB open off-thread, but the foreground path still waits for it. A follow-up could make `ready()` complete asynchronously after init. Called out here for visibility; out of scope for this regression fix.

## Testing

Not compiled locally (native + Rust toolchain); verified by inspection, as with #252/#260. Repro: `startOnBoot: true` + `stopOnTerminate: false`, cold boot / task-removal restart.

### DIFF
--- a/sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt
+++ b/sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt
@@ -291,7 +291,34 @@ class TraceletSdk private constructor(private val context: Context) {
         check(::eventSender.isInitialized) {
             "setEventSender() must be called before initialize()"
         }
+        // Run the heavy setup — which opens the Rust DB and fsyncs it to disk —
+        // off the calling thread. onAttachedToEngine runs on the platform main
+        // thread; when a background FlutterEngine (e.g. audio_service's
+        // MediaBrowserService, spun up by the system after the app was killed)
+        // attaches, GeneratedPluginRegistrant re-attaches this plugin and this
+        // init would otherwise fsync on that service's main thread → ANR on a
+        // large DB. ready() blocks on initCompleteLatch before it uses the DB or
+        // the engines set up here.
+        synchronized(initLock) {
+            if (initStarted) return
+            initStarted = true
+        }
+        Thread({
+            try {
+                initializeInternal()
+            } catch (t: Throwable) {
+                logger.error("initializeInternal failed: ${t.message}")
+            } finally {
+                initCompleteLatch.countDown()
+            }
+        }, "tracelet-init").start()
+    }
 
+    private val initLock = Any()
+    @Volatile private var initStarted = false
+    private val initCompleteLatch = java.util.concurrent.CountDownLatch(1)
+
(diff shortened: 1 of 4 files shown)


GalacticTitan:
Superseded by #271, which is merged to `main`. Your original commits are preserved as ancestors of the merge — I rebased/merged them onto post-#264 `main`, resolved the conflicts (kept `main`'s `initializationFailure: Throwable?` and the defensive `lateinit` re-check), routed the two broadcast receivers through the new `awaitInit()` gate, and added unit tests + an example repro card. Thanks for the fix! 🙏

Issue #274, opened 2026-07-29
Geofence: GPS drift causes false EXIT events on small-radius geofences


GalacticTitan:
## Summary

In `geofenceModeHighAccuracy`, a stationary device inside a geofence can emit a **false EXIT** when a single GPS fix drifts beyond the radius, even though the user has not moved. This is especially problematic for small-radius attendance geofences (e.g. 50 m).

## Steps to reproduce

1. Register a small circular geofence (e.g. radius 50 m).
2. Arrive at the location → **ENTER** fires correctly.
3. Walk away → **EXIT** fires correctly.
4. Stand still well inside the fence. Occasionally GPS reports a high-drift fix (large horizontal accuracy error) that lands outside `radius + hysteresis buffer`.
5. **A false EXIT fires** even though the device never moved.

## Root cause

`GeofenceEvaluator::evaluate_proximity` (Rust core, mirrored in Dart) decides EXIT purely on `distance > radius + exitBuffer`. The existing exit-hysteresis band (#268) absorbs small stationary jitter, `GeofenceEv single-fix drift spike sails past it.

Critically, the evaluator **ignores the fix's reportCritically, the evaacy`*Criticallreported as "here, ±150 m" is trusted exactly as much as one reported "±5 m". `LocationEngine.onLocationUpdate` drops accuracy `(lat, lng)` before it ever reaches the evaluator.

## Proposed fix (this issue)

Make the EXIT decision **accuracy-aware**: only fire EXIT once the entire GPS error circle clears the fence:

```
exited = (distance - horizontalAccuracy) > radius + exitBuffer
```

A ±150 m fix reported 160 m out yields `160 - 150 = 10 m` → still "inside", no false EXIT. A genuine departure with a good ±8 m fix at 90 m yields `82 m` → real EXIT. ENTER behavior is unchanged. This is always-on and requires no new public API.

Scope:
- Rust core `geofence_evaluator.rs` (canonical) + regenerated UniFFI bindings.
- Native `GeofenceManager.evaluateHighAccuracyProximity` + `LocationEngine.onLocationUpdate` wiring (Android + iOS) to thread accuracy through.
- Pure-Dart `geofence_evaluator.dart` mirror.

## Follow-up (not in this issue)

Expose tunable knobs on `GeofenceConfig` for teams that want stricter control:
- `geofenceMaxAccuracyRadius` — ignore fixes worse than N metres for EXIT decisions.
[...]

PR #281, opened 2026-07-29
fix(sdk): persist locationSource/reducedAccuracy so DB reads & sync keep them (#280)

### LINKED ISSUE #280: [Bug]: locationSource / reducedAccuracy dropped from persisted & synced locations (always "unknown"/false)
## Summary

`locationSource` and `reducedAccuracy` are present on the **live** `onLocation` event but are **silently dropped** for every DB-sourced read — `getLocations()`, `getPendingLocations()`, and the DB-sourced sync payload (`setSyncBodyBuilder`). They always come back as `locationSource: "unknown"` and `reducedAccuracy: false`.

## Root cause

The live map builder emits both fields at the **top level** of the location map:
- iOS: `LocationEngine.buildLocationMap` (`sdk/ios/.../location/LocationEngine.swift`)
- Android: `LocationEngine` (`sdk/android/.../location/LocationEngine.kt`)

But the persist path only serializes `audit_*`, `battery`, and `extras` into the `route_context` column:
- iOS: `TraceletSdk.insertLocation` (`sdk/ios/.../TraceletSdk.swift`)
- Android: `TraceletSdk.insertLocation` (`sdk/android/.../TraceletSdk.kt`)

The top-level `locationSource`/`reducedAccuracy` are never written, so they are lost at persist time. On read, `LocationMapper.buildLocationMap` (both platforms) never restores them, and `Location.fromMap` falls back to the defaults.

## Impact

`help/FAQ.md` explicitly tells users to filter by `locationSource == "gps"` (or `accuracy <= 50`) to distinguish GPS from Wi-Fi/cell fixes. That guidance silently fails for any historical or synced record because the tag is gone — every persisted record reads "unknown", and `reducedAccuracy` (used to detect coarse/Precise-Off fixes) is always false.

## Expected

`locationSource` and `reducedAccuracy` should survive persistence and be reported identically by the live event and DB-sourced reads/sync.

## Fix

Serialize both as first-class `route_context` keys on persist (symmetric with `audit_hash`) and promote them back to top level in `LocationMapper` on both platforms, so they never leak into `extras`. Includes unit tests and an example repro card.

Found while auditing a downstream fork; reworked to avoid stuffing the fields into `extras`.

GalacticTitan:
Closes #280.

## Problem

`locationSource` and `reducedAccuracy` are emitted on the **live** `onLocation` event (top-level keys set by `LocationEngine.buildLocationMap` on both platforms), but they are **dropped when persisted**. `insertLocation` only serializes `audit_*`, `battery` and `extras` into the `route_context` column, so:

- `getLocations()` / `getPendingLocations()` always return `locationSource: "unknown"`, `reducedAccuracy: false`.
- The DB-sourced sync payload (`setSyncBodyBuilder`) carries the same lost values.

This breaks `help/FAQ.md`'s documented guidance to filter historical/synced fixes by `locationSource == "gps"`, and makes `reducedAccuracy` (coarse / Precise-Off detection) useless off the live stream.

## Fix

- **Persist** `locationSource` / `reducedAccuracy` as first-class `route_context` keys (symmetric with `audit_hash`), only when present — no behavior change for records that don't carry them.
- **Restore** them in `LocationMapper.buildLocationMap` by promoting them to the top level, where `Location.fromMap` reads them.
- Kept **out of `extras`** so they don't leak into `extras.route_context` (cleaner than the naive approach of stuffing them into extras).

Applied on both Android (`TraceletSdk.kt`, `LocationMapper.kt`) and iOS (`TraceletSdk.swift`, `LocationMapper.swift`).

## Tests

- `LocationMapperTest.kt` / `LocationMapperTests.swift`: assert the two fields promote to top level and do **not** leak into `extras.route_context`, plus an absence case.
- Example app: new on-device E2E repro card (#280) that inserts a tagged location, persists, reads it back via `getLocations()`, and asserts the tags survived.

## Verification

- Android `:tracelet-sdk:testDebugUnitTest --tests LocationMapperTest` — **passes** (new cases included), main source compiles.
- `melos run format` + `melos run analyze` — **clean**.
- iOS: `LocationMapper.swift` / `TraceletSdk.swift` parse clean; the full iOS test target could not be built in my environment due to pre-existing generated-FFI (`tracelet_core.swift`) drift unrelated to this change — CI should run the iOS suite. The iOS logic mirrors the Android change 1:1 (verified by the shared `LocationMapperTests`).

## Background

Surfaced while auditing a downstream fork that had a similar fix; this PR is a cleaner, from-scratch reimplementation against current `main` (no `extras` leak, guarded serialization, tests on both platforms).

### DIFF
--- a/example/lib/issues/issue_280_card.dart
+++ b/example/lib/issues/issue_280_card.dart
@@ -1,0 +1,111 @@
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+
+/// Issue #280 — `locationSource` / `reducedAccuracy` dropped from persisted &
+/// synced locations.
+///
+/// Both fields are emitted on the live `onLocation` event at the top level of
+/// the location map, but the persist path only serialized `audit_*`, `battery`
+/// and `extras` into the `route_context` column — so the classification was
+/// lost at write time and every DB-sourced read (`getLocations`,
+/// `getPendingLocations`, the DB-sourced sync payload) reported
+/// `locationSource: "unknown"` / `reducedAccuracy: false`.
+///
+/// The fix persists both as first-class `route_context` keys (like `audit_*`)
+/// and `LocationMapper` promotes them back to the top level on read, so the
+/// live event and DB-sourced reads agree.
+///
+/// This card is an on-device E2E: it inserts a location tagged
+/// `locationSource: "gps"`, `reducedAccuracy: true`, persists it, reads it back
+/// via `getLocations()`, and asserts the tags survived the round-trip.
+class Issue280Card extends StatefulWidget {
+  const Issue280Card({super.key});
+
+  @override
+  State<Issue280Card> createState() => _Issue280CardState();
+}
+
+class _Issue280CardState extends State<Issue280Card> {
+  String _status = 'Idle';
+  bool _running = false;
+
+  void _set(String s) {
+    if (mounted) setState(() => _status = s);
+  }
+
+  Future<void> _run() async {
+    setState(() => _running = true);
+    try {
+      _set('Testing #280 (persist locationSource / reducedAccuracy)...');
+      await Tracelet.ready(const Config());
+      await Tracelet.destroyLocations();
+
+      const uuid = 'issue-280-gps';
+      await Tracelet.insertLocation({
+        'uuid': uuid,
+        'timestamp': DateTime.now().toUtc().toIso8601String(),
+        'coords': {
+          'latitude': 37.7749,
+          'longitude': -122.4194,
+          'accuracy': 8.0,
+        },
+        // The classification the live onLocation event carries — must survive
+        // persistence and come back on the DB-sourced read.
+        'locationSource': 'gps',
+        'reducedAccuracy': true,
+      });
+
+      final locations = await Tracelet.getLocations();
+      if (locations.isEmpty) {
+        _set('❌ FAILED: no locations returned from getLocations().');
+        return;
+      }
+      final loc = locations.firstWhere(
+        (l) => l.uuid == uuid,
+        orElse: () => locations.first,
+      );
+
+      final sourceOk = loc.locationSource == 'gps';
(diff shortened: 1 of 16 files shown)


GalacticTitan:
### On-device card failure investigated — no code regression

The #280 example card was reported failing on-device (`locationSource="unknown"`). I traced the full Android path and added three layers of automated proof, all green:

1. **LocationMapper unit tests** — promotion to top level, no `extras` leak.
2. **DB round-trip** (`TraceletDatabaseQueryTest`, real Rust `DatabaseManager`) — a `route_context` carrying `locationSource`/`reducedAccuracy` persists and is promoted on read.
3. **Full write path** (`TraceletSdkLocationSourcePersistenceTest`, Robolectric + real DB) — `TraceletSdk.insertLocation({... top-level locationSource/reducedAccuracy ...})` → `getLocations()` returns `locationSource="gps"`, `reducedAccuracy=true`.

So the persist → read → map chain is correct end-to-end. The on-device failure was a **stale native build**: Flutter hot-reloads the new Dart card, but the native Kotlin SDK change is only picked up on a native recompile. A `flutter clean` (or full rebuild) of the example app resolves it — the composite build (`includeBuild ../../sdk/android`) then compiles the patched SDK source.

Note: the 2 `GeofenceManagerExitAccuracyTest` failures in the full suite are **pre-existing on `main`** (unrelated local Rust-core drift), not from this change.

GalacticTitan:
### iOS side now verified

Rebuilt the stale local `TraceletCore.xcframework` (its embedded FFI header predated the #274 `accuracy` param, which blocked `xcodebuild`). With that fixed, I also wired `LocationMapperTests.swift` into the SwiftPM test target (commit d6643d8) — it was previously shelved but only uses the stable `buildLocationMap` API.

Result on the iOS simulator: the two #280 cases pass (`...LocationSourceAndReducedAccuracyGoTopLevelNotIntoExtras`, `...WithoutLocationSourceLeavesKeysAbsent`), and the full `TraceletSDKTests` suite is green (34 tests, 0 failures). So #280 is now verified on **both** Android (unit + DB round-trip + full SDK write path) and iOS.

PR #283, opened 2026-07-29
fix(ios): high-accuracy periodic fix uses best-of-N sampling window (#282)

### LINKED ISSUE #282: [Enhancement]: iOS periodic fix uses a single requestLocation() (can persist a stale/coarse fix) instead of the best-of-N window
## Summary

On iOS, `LocationEngine.performPeriodicFix()` (used by periodic / stationary-periodic tracking) takes a single `CLLocationManager.requestLocation()` one-shot. Per the SDK's own note in `collectSamples`, `requestLocation()` "may return a stale cached location without waking the GPS hardware," and even a fresh one-shot is often the first coarse fix before GPS converges. The result is that periodic fixes are frequently lower-accuracy (Wi-Fi/cell-level) than they need to be.

## Existing infrastructure

`getCurrentPosition` already solves exactly this with a best-of-N sampling window that is well-tested and integrated:

- `collectSamples(count:persist:extras:callback:)` — `startUpdatingLocation()` at `kCLLocationAccuracyBest`, no distance filter, bounded by `locationTimeout`.
- `feedSample(_:)` — `didUpdateLocations` routes samples here and gates normal processing on `consumedBySampler`, so there's no double-dispatch.
- `deliverBest(...)` — selects the minimum-`horizontalAccuracy` sample, persists + dispatches it.
- `restoreAfterSampling()` — already handles the periodic case, restoring the low-power periodic configuration afterwards.

`performPeriodicFix` does **not** use any of this today.

## Proposal

When periodic desired accuracy is 'best' (`getPeriodicDesiredAccuracy() == 0`), route `performPeriodicFix()` through `collectSamples(count: N, persist: true)` (wrapped in the existing periodic background task) instead of `requestLocation()`. Non-best periodic accuracy keeps the current single-shot path. This reuses the verified sampling/selection/restore logic and yields materially better periodic fixes for a bounded extra GPS window.

## Notes / tradeoffs

- Slightly more GPS-on time per fix (the sampling window, capped by `locationTimeout`) in exchange for accuracy — only when accuracy is 'best'.
- No double-dispatch: `feedSample` consumes samples during the window.
- Needs on-device / CI verification of power + indicator behavior.

Surfaced while auditing a downstream fork that reinvented this with a parallel high-accuracy sampling path; the upstream-appropriate approach is to reuse the existing `collectSamples` window.

GalacticTitan:
Closes #282.

## Problem

On iOS, `LocationEngine.performPeriodicFix()` takes a single `CLLocationManager.requestLocation()` one-shot. Per the SDK's own note in `collectSamples`, `requestLocation()` may return a stale cached location without waking the GPS hardware, and even a fresh one-shot is often the first coarse fix before GPS converges — so a periodic tick can persist a Wi-Fi/cell-level fix when a GPS-quality one was achievable.

## Fix

When periodic desired accuracy is `DesiredAccuracy.high` (native `kCLLocationAccuracyBest`), route `performPeriodicFix()` through the existing best-of-N sampling window that `getCurrentPosition` already uses:

- `collectSamples(count: 3, persist: true)` — `startUpdatingLocation()` at best accuracy, bounded by `locationTimeout`.
- `feedSample()` consumes the samples during the window, so there is **no double-dispatch** (`didUpdateLocations` gates normal processing on `consumedBySampler`).
- `deliverBest()` persists + dispatches the minimum-`horizontalAccuracy` sample.
- `restoreAfterSampling()` restores the low-power periodic configuration afterwards.

Non-high periodic accuracy keeps the cheaper single `requestLocation()` path unchanged. An `in-flight window` guard (`sampleState == nil`) prevents overlapping windows if the periodic interval is shorter than `locationTimeout`.

## Tradeoffs

- Slightly more GPS-on time per fix (the sampling window, capped by `locationTimeout`) in exchange for GPS-quality periodic fixes — only when accuracy is high.
- Reuses already-shipped, `getCurrentPosition`-tested infrastructure rather than adding a parallel path.

## Verification

- `melos run format` + `melos run analyze` — **clean** (includes the new example card).
- `LocationEngine.swift` parses clean.
- **Not yet run on-device / in CI here**: the full iOS test target could not be built in my environment due to pre-existing generated-FFI (`tracelet_core.swift`) drift unrelated to this change. This touches a core periodic path, so please verify periodic power/indicator behavior on a device (or via CI). A manual on-device iOS repro card (#282) is included that lists periodic fix accuracy / `locationSource`.

## Background

Surfaced while auditing a downstream fork that reinvented this with a parallel high-accuracy sampling path; this PR instead reuses the existing `collectSamples` window against current `main`.


GalacticTitan:
### iOS build/verification update

I rebuilt the local Rust core artifact (the gitignored `TraceletCore.xcframework` in my checkout was stale — its embedded FFI header predated the #274 `evaluate_proximity(accuracy)` change, which is what blocked `xcodebuild` earlier). With a fresh `build-ios.sh`, the iOS build works.

On this branch the `LocationEngine.swift` periodic change **compiles cleanly** and the full `TraceletSDKTests` suite is green on the iOS simulator (22 tests, 0 failures), including `LocationEngineRuntimeProviderOptionsTests`. There is still no dedicated automated test for the periodic timing/power behavior, so on-device confirmation of the best-of-N window is still recommended, but the change is confirmed to build and not regress the existing iOS suite.

Issue #213, opened 2026-06-18
[Bug]: Background sync continues sending payloads after calling Tracelet.stop() (isEnabled = false)


scalz:
### Package Name

tracelet (Core Plugin)

### Tracelet Version

3.4.1

### Platform

Android

### Device & OS Version

Emulator Android 16

### Description / Scenario

When stopping the tracking engine by calling Tracelet.stop() (which correctly updates the SDK state to enabled: false and stops GPS polling), any debounced HTTP auto-sync task that is currently waiting out its delay (configured via HttpConfig.autoSyncDelay, defaulting to 10 seconds) is not cancelled.

As a result, the background worker continues to execute its network request and transmits the remaining location queue to the backend server up to 10+ seconds after the user has explicitly stopped geolocation. This creates unintended background data usage and violates user expectations regarding immediate tracking cessation.

**Probable root cause**
During TraceletSdk.stop() execution in TraceletSdk.kt, the SDK attempts to cancel any pending synchronization jobs by calling:
```
syncProvider?.cancelPendingSync()
```

The SyncProvider interface defines this as a default no-op method. While this method is properly overridden and implemented inside the native-only NativeSyncProvider.kt fallback class, it was omitted from TraceletSyncSink inside TraceletSyncPlugin.kt (which is registered as the active sync provider when the Flutter engine is attached and running in the foreground).

Because TraceletSyncSink does not override this method, calling cancelPendingSync() does nothing. The active coroutine syncJob (launched in insertLocation) continues to sleep through its delay(delayMs) call, and eventually executes the HTTP sync.

**Proposed Resolution (untested)**

To resolve this issue, the TraceletSyncSink class inside the tracelet_sync package must override and implement cancelPendingSync() to cancel the active coroutine job.

In packages/tracelet_sync/android/src/main/kotlin/com/ikolvi/tracelet_sync/TraceletSyncPlugin.kt:
```
class TraceletSyncSink(private val sdk: TraceletSdk) : LocationDataSink, TraceletSdk.SyncProvider {
    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
    private val syncMutex = Mutex()
    private val syncManager = SyncManager()

    private var syncJob: kotlinx.coroutines.Job? = null
    private val DEBOUNCE_MS = 10_000L

    override fun insertLocation(location: Map<String, Any?>) {
[...]

Issue #204, opened 2026-06-16
[Bug]: requestSyncBody invoked multiple times for same batch causing duplicate server uploads


Srj-Love:
### Package Name

tracelet (Core Plugin)

### Tracelet Version

3.3.4

### Platform

Android

### Device & OS Version

Samsung

### Description / Scenario

We are observing duplicate GPS records being uploaded to our backend because `requestSyncBody` is being invoked multiple times for the exact same batch of locations.

From the application logs, Tracelet calls the interceptor twice with the same batch, resulting in two identical HTTP uploads and duplicate records in the database.

---

## Expected Behavior

For a given batch of locations, `requestSyncBody` should be invoked only once.

Example:

```text
Calling requestSyncBody on interceptor with 3 records
requestSyncBody called with 3 locations
executeFallbackHttpSync success: true
Synced and cleared 3 locations
```

Only one invocation should occur per batch.

---

## Actual Behavior

The same batch is processed twice.

Logs show:

```text
D/Tracelet: Calling requestSyncBody on interceptor with 3 records
D/Tracelet: Calling requestSyncBody on interceptor with 3 records

D/Tracelet: requestSyncBody called with 3 locations. isEngineAttached=true
D/Tracelet: requestSyncBody called with 3 locations. isEngineAttached=true
```

The Flutter callback executes twice for the same location batch.

Subsequently:

```text
executeFallbackHttpSync success: true
executeFallbackHttpSync success: true

Synced and cleared 3 locations.
Synced and cleared 3 locations.
```

This results in duplicate API requests and duplicate database rows.

---

### Duplicate interceptor invocation

```text
Calling requestSyncBody on interceptor with 3 records
Calling requestSyncBody on interceptor with 3 records
```

### Duplicate callback execution

```text
requestSyncBody called with 3 locations. isEngineAttached=true
requestSyncBody called with 3 locations. isEngineAttached=true
```

### Duplicate sync completion

```text
executeFallbackHttpSync success: true
executeFallbackHttpSync success: true

Synced and cleared 3 locations.
Synced and cleared 3 locations.
```

---

## Backend Impact

Example duplicate rows inserted:

```text
Jun 16, 2026, 2:01:35 PM    19.1664004018216    73.00112969937787
Jun 16, 2026, 2:01:35 PM    19.1664004018216    73.00112969937787
```

Both rows originate from the same GPS record.

---

## Additional Notes

Our interceptor implementation only generates the request body and does not perform any retry logic.

[...]

Issue #230, opened 2026-06-30
[Bug]: Runtime configuration changes (setConfig) do not propagate to active native tracking/sensor loops


scalz:
### Package Name

tracelet (Core Plugin)

### Tracelet Version

3.5.3

### Platform

Android

### Device & OS Version

Emulator Android 17, Pixel 8 Pro Android 17

### Description / Scenario

This has been reported by one of my users. So I tried to understand where comes from this behaviour. Maybe I could use a workaround by doing a stop, delay, start, when calling setConfig at runtime but I've not tried yet, also I'm planning to call setConfig from headless task when some event trigger (eg. for switching profile on event, even when app is killed).

When a client updates the SDK configuration at runtime using Tracelet.setConfig(config), the new settings are saved to ConfigManager and synchronized to the Rust Core flat config (syncConfigToRustFlat()).

However, the active native tracking loops do not receive these updates unless the app is forcefully killed and restarted.

In TraceletSdk.kt, the config change re-evaluation is constrained to this block:
```
if (stateManager.enabled) {
    val locationKeys = listOf(
        "desiredAccuracy", "distanceFilter", "locationUpdateInterval",
        "fastestLocationUpdateInterval", "stationaryRadius", "deferTime",
        "disableElasticity", "elasticityMultiplier"
    )
    val needsRestart = locationKeys.any { key -> oldConfig[key] != merged[key] }
    if (needsRestart) {
        locationEngine.stop()
        locationEngine.start()
    }
}
```

This introduces two points of failure:
1. Motion Configuration Ignored: Sane motion-detector or speed-manager configurations—such as motionDetectionMode (e.g., switching from ACCELEROMETER to SPEED), shakeThreshold, speedMovingThreshold, or stopTimeout—are omitted from locationKeys. Changes to these properties never trigger needsRestart.

2. Subsystems Never Rebuilt: Even when needsRestart is true, the code only restarts locationEngine. It completely neglects to stop and recreate motionDetector, speedMotionManager, or smartMotionCoordinator. Consequently, the physical hardware sensors and speed state machines continue running on the device with their old parameters, leading to an inconsistent tracking state.

#### Impact

Critical configuration adjustments (like switching motion detection strategies, raising shake thresholds, or shortening stationary timeouts) are ignored at runtime. The user is forced to kill the app from Android settings and relaunch it to apply the new settings.

#### Proposed Resolution (untested)

[...]

PR #287, opened 2026-08-01
fix: process-wide sync sink + speed-noise stationary stall (#286)

### LINKED ISSUE #286: Per-engine TraceletSyncSink instances accumulate → concurrent auto-syncs → OOM (thread + heap)
# Per-engine `TraceletSyncSink` instances accumulate → concurrent auto-syncs → OOM (thread + heap)

**Packages:** `tracelet_sync` 3.6.14, `tracelet` / `tracelet_android` 3.6.14 (also present in 3.6.6)
**Platform:** Android (Flutter 3.x, `workmanager_android` 0.9.0+2 also in the app)
**Severity:** fatal — three distinct crash signatures in production, frequency grows with process uptime

## Summary

`TraceletSyncPlugin.onAttachedToEngine` creates a **new** `TraceletSyncSink` for every `FlutterEngine`
that attaches, and `onDetachedFromEngine` never tears it down. Any host app that spawns secondary
Flutter engines (WorkManager background tasks, headless services, engine groups) therefore accumulates
sinks over time. Each surviving sink keeps its **own** `CoroutineScope`, `Mutex` and `SyncManager`, so
the per-sink concurrency guards (`syncJob?.isActive` and `syncMutex`) no longer serialize anything:
every persisted location fans out into N blocking auto-syncs.

```kotlin
// TraceletSyncPlugin.kt:311-321
override fun onAttachedToEngine(binding: FlutterPlugin.FlutterPluginBinding) {
    ...
    val sink = TraceletSyncSink(traceletSdk)   // :319  new instance per engine
    traceletSdk.registerSyncProvider(sink)     // :320
}

// TraceletSyncPlugin.kt:338-340
override fun onDetachedFromEngine(binding: FlutterPlugin.FlutterPluginBinding) {
    channel.setMethodCallHandler(null)         // :339  sink not cancelled, scope not closed
}
```

`TraceletSdk.registerSyncProvider` does try to unregister the previous provider from the location
engine, but that path is gated on `locationEngine != null` at the moment of registration, so in
practice stale sinks stay subscribed and keep receiving `insertLocation`.

## Evidence

Three production Crashlytics reports from the same app. In each one, several threads are inside
`TraceletSyncSink.triggerSync`, all entered from `insertLocation$1.invokeSuspend
(TraceletSyncPlugin.kt:38)`. Because a single sink serializes `triggerSync` behind `syncMutex`
(TraceletSyncPlugin.kt:24), **k concurrent `triggerSync` frames imply k live sink instances.**

**Crash 1 — `OutOfMemoryError: pthread_create (1040KB stack) failed`**
9 concurrent syncs. Each blocking sync pins 1–2 threads:
`syncBatchBlocking` → `runBlocking` (:165) plus the nested `executeFallbackHttpSync` (:268), or the
uniffi `syncManager.syncBatchBlocking` (:193), or `requestSyncBody` waiting on the Dart latch (:150).
Thread limit exhausted; the process died in an unrelated component that happened to ask for a thread.

**Crash 3 — `OutOfMemoryError: Failed to allocate ... target footprint 268435456`**
6 concurrent syncs; three of them blocked in
`HeadlessTaskService.requestCustomSyncBody (HeadlessTaskService.kt:245)`, two lifting batches out of
the Rust buffer in `DatabaseManager.getLocationsBatch (TraceletSyncPlugin.kt:76)`. The coroutine
scheduler had grown to 23 workers.

## Reproduction

1. App with `tracelet` + `tracelet_sync` and `http.autoSync = true` on Android.
2. Any component that creates additional Flutter engines. `workmanager_android` does this per task:
   `BackgroundWorker.kt:68` `engine = FlutterEngine(applicationContext)` … `:187 engine?.destroy()`.
3. Start tracking and let the app run for hours while background tasks fire periodically.
4. Log/count `"Sync sink registered!"` (TraceletSyncPlugin.kt:323) — it grows monotonically, and the
   number of `"triggerSync started"` lines per location interval grows with it.

Side effects before the crash: the same batch is POSTed by several sinks (duplicate points server
side), and concurrent `clearLocationsUpTo` calls race on the same rows.

## Suggested fix

Either of:

- Make the sink **process-wide**: keep it in a companion object, create it on first attach, and reuse
  it for every subsequent engine — the SDK is already a singleton (`TraceletSdk.getInstance`).
- Or tear the sink down in `onDetachedFromEngine`: `syncSink?.cancelPendingSync()` plus cancelling its
  `CoroutineScope` (`scope.cancel()`) and unregistering it as a `LocationDataSink`.

Two smaller issues found while investigating, both amplifying the impact:

1. `executeFallbackHttpSync` (TraceletSyncPlugin.kt:~296-300) hardcodes
   `connectTimeout = 15000` / `readTimeout = 15000` and ignores the configured `HttpConfig.httpTimeout`.
   With `maxRetries = 3` a single sync can hold two threads for ~2 minutes.
2. `syncBatchBlocking` uses `runBlocking` inside a `Dispatchers.IO` coroutine (:165), so every in-flight
   sync costs two threads instead of one. A suspending HTTP path would remove the thread doubling.

## Workarounds currently applied in the app

- `maxRetries: 3 → 1` and `autoSyncDelay: 12s → 30s` — reduces how many blocking syncs overlap.
- Live points are pushed by a native keep-alive loop instead of relying on auto-sync latency.

These reduce the crash rate but cannot stop the accumulation itself.

### LINKED ISSUE #288: Pace never leaves MOVING in smart motion mode (or takes minutes on Android)
# Pace never leaves MOVING in smart motion mode (or takes many minutes on Android)

**Packages:** `tracelet` / `tracelet_android` / `tracelet_ios` 3.7.1 (present since 3.2.x)
**Platform:** both; far more visible on Android
**Severity:** high — continuous GPS never steps down to periodic/stationary, so battery drain continues indefinitely and `isMoving` stays wrong

## Summary

With `motionDetectionMode: MotionDetectionMode.smart`, a device that is physically still often never transitions to STATIONARY. Forcing it with `changePace(false)` always works, which makes it look like the automatic path is dead. On Android it can take many `speedStationaryDelay` windows or never complete; on iOS it usually completes but for unrelated reasons (cleaner `CLLocation.speed`).

Three independent defects combine.

## 1. A single noisy GPS speed sample discards the whole SLOWING countdown

`SpeedMotionManager.onLocationSlowing` (Android) / `handleSlowing` (iOS) treats one fix at or above `speedMovingThreshold` as "moving again": it cancels the timer and returns to MOVING, so the next low fix restarts the full `speedStationaryDelay` window.

GPS speed is noisy on a stationary device. From a device log with `speedStationaryDelay: 30`:

```
15:30:36.058  State transition: MOVING -> SLOWING
15:30:38.672  [STILLNESS] ★★★ sustained stillness detected (25 samples)
15:30:47.749  SLOWING: lowCount=3, elapsed=11692ms, delay=30000ms, speed=0.34 m/s
15:30:49.781  SLOWING: lowCount=4, elapsed=13724ms, delay=30000ms, speed=0.00 m/s
   ... all subsequent fixes 0.00 m/s ...
15:30:58.850  State transition: SLOWING -> MOVING     ← one fix at 1.56 m/s, 21.7s into 30s
15:31:01.741  State transition: MOVING -> SLOWING     ← countdown restarts from zero
```

The accelerometer had already reported sustained stillness. One blip amid a stream of `0.00 m/s` fixes threw away 21.7 s of progress, and this repeats indefinitely.

`MotionDetector` already solved exactly this for accelerometer noise in `daf61dc8` (`MOTION_ABORT_COUNT`, requiring *sustained* motion to abort a countdown). The speed machine never got the same treatment.

## 2. The accelerometer input is inert after a start that begins in MOVING

The Rust `SmartMotionCoordinator` initialises `is_accel_moving = false`, and `on_accel_state_change()` early-returns when the flag is unchanged. `start()` never seeded it. So when `MotionDetector`'s `stopTimeout` fires and reports stationary, the coordinator sees no transition and emits no action — the GPS-speed machine is the only input that can ever change the pace.

This is why defect 1 is fatal rather than merely slow: the accelerometer cannot compensate.

## 3. (Android) the coordinator's tracking mode is only synced in `initialize()`

`smartMotionCoordinator.syncCurrentMode()` runs in `initialize()` from the **persisted** `trackingMode`, while `start()` sets `trackingMode = CONTINUOUS` without re-syncing. A session that ended stationary therefore leaves the Rust coordinator believing it is already in `StationaryPeriodic`, and `evaluate_state()` only emits a stationary action while it believes the mode is `Continuous`. The pace then cannot leave MOVING until the process is killed.

Manual `changePace()` still works in all three cases because it sets *both* coordinator flags.

## Also found while auditing the same area

`MotionDetector.declareStationary()` writes `state.isMoving = false` even in smart mode, where the coordinator owns the decision — so `getState().isMoving` can disagree with the last `onMotionChange` event and with actual GPS behaviour.

Sensor thresholds are a single cross-platform scalar whose Dart defaults are the Android values, and they were sent unconditionally by any app that configured *any* motion field. iOS converts m/s² to g, so `stillThreshold: 0.4` arrived as `0.04 g` — roughly four times stricter than the `0.15 g` iOS default it was meant to keep — and `stillSampleCount: 25` dwelt ~2.5 s at iOS's 10 Hz instead of the intended ~5 s. iOS's own `getStillSampleCount()` fallback (30) also disagreed with the documented 50, and three `private static let` constants in the iOS `MotionDetector` documented the tuned values but were never referenced.

## Reproduction

1. `motionDetectionMode: MotionDetectionMode.smart`, `stopTimeout: 1`, `speedStationaryDelay: 30` (the example app's config).
2. Start tracking, put the phone on a desk, do not touch it.
3. Watch for `State transition: SLOWING -> MOVING` lines whose reported speed is barely above `speedMovingThreshold` while neighbouring fixes read `0.00 m/s`. Each one restarts the countdown.
4. For defect 3: end a session in the stationary state, then start a new one — the pace never leaves MOVING.

## Not a regression

The single-sample abort is original code from `5421e7a0 feat: Speed-Based Motion Detection (#83)`. It became user-visible when `41294020` switched the example (and the recommended setup) from `MotionDetectionMode.accelerometer` to `.smart`: in accelerometer mode the pace flipped purely on `MotionDetector`, so GPS speed noise never entered the decision.

## Suggested fix

- Require several consecutive above-threshold fixes before abandoning a SLOWING countdown, and keep the countdown's original start time across an isolated blip. Safe because SLOWING is still continuous tracking, so confirming over a couple of fixes costs no location fidelity.
- Seed the coordinator's accelerometer flag at `start()` from the state being started in, and re-sync its tracking mode there.
- Let the coordinator own `isMoving` in smart mode.
- Only transmit sensor thresholds the app actually set, so each platform keeps its tuned default.

GalacticTitan:
Fixes #286.

## Summary

Both native `tracelet_sync` plugins built a **new** `TraceletSyncSink` for every `FlutterEngine` that attached, and neither ever detached one. Hosts that spawn secondary engines — `workmanager` creates one per background task, plus headless engines and engine groups — accumulated sinks for the life of the process.

Each sink owns its own concurrency guard: a `CoroutineScope` + `Mutex` on Android, a `SyncCoordinator` actor on iOS. Once several exist, those guards can only serialize their own instance, so one persisted location fans out into N blocking auto-syncs, each pinning one or two threads. That is the reporter's three production signatures: `OutOfMemoryError: pthread_create (1040KB stack) failed`, heap exhaustion at a 256 MB footprint, plus duplicate points server-side and racing `clearLocationsUpTo` calls.

`TraceletSdk` is a process singleton and only one sync provider can be active, so the sink is now created once and reused by every later engine. That is the first option the issue suggests, and it keeps working for headless/native tracking: the sink is deliberately *not* torn down on detach, because a short-lived background engine going away must not stop auto-sync for the rest of the process. `TraceletSdk.stop()` still cancels the pending debounce (#213).

**iOS was affected too, with fewer brakes than Android** — no replace-and-unregister step, no `unregisterSink` at all, no detach hook, and the plugin subscribed the same sink twice per engine (directly *and* through the `syncProvider` `didSet`, where `registerSink` was a bare `append`).

## Changes

**Android — `packages/tracelet_sync`**

- sink held in a `@Volatile` companion field, created under a lock on first attach (double-checked locking: a background engine can attach while the UI engine still is)
- `onAttachedToEngine` reuses it and re-registers it idempotently, so a newly built `LocationEngine` still gets it
- documented why `onDetachedFromEngine` must not tear it down

**iOS — `packages/tracelet_sync`**

- same process-wide sink behind an `NSLock`
- `register(with:)` no longer subscribes the sink directly *and* assigns `syncProvider`; the `didSet` does the subscribing, so the sink stopped being added twice per engine and registration no longer depends on `locationEngine` already being non-nil
- added `detachFromEngine(for:)` to clear the channel handler

**iOS SDK — parity with Android**

- `LocationEngine.registerSink` dedupes by identity; `unregisterSink` added (iOS had no way to detach a sink at all)
- `TraceletSdk.syncProvider` `didSet` cancels and unregisters a superseded provider, as Android's `registerSyncProvider` has since #204
- `sinks` is now `internal private(set)` so tests can assert what is subscribed

## Tests

- `packages/tracelet_sync/test/issue_286_sink_accumulation_test.dart` — 10 CI-gated source invariants across both platforms: the sink is constructed in exactly one place (the shared accessor) and never in the per-engine entry point, is held statically, is subscribed once on iOS, the iOS plugin has a detach hook, and both SDKs can detach a sink. The sinks and their scopes are native internals with no Dart-visible API, and reproducing the accumulation needs real secondary engines, so these guard the invariants that make it impossible.
- `sdk/ios/Tests/TraceletSDKTests/Issue286SinkAccumulationTests.swift` — `registerSink` dedupe by identity, distinct sinks both kept, `unregisterSink` removes only its own sink, unknown sink is a no-op.

## Verification

- **Reproduced on-device before the fix** (Android debug build, via the new example card): 2 attached-then-destroyed engines produced 3 distinct sinks, all 3 with a live `CoroutineScope` after `engine.destroy()`, and the SDK left pointing at a sink from a destroyed engine.
- The example app gained a `#286` card (Recent Issues → search `286` → Verify) that attaches and releases secondary engines on device and reports distinct sinks, survivors and `LocationEngine` subscriptions, on **both** platforms. It verifies only; it does not fix.
- `melos run format` and `melos run analyze` clean; `flutter test` green in `packages/tracelet_sync`; `flutter build apk --debug`, `:tracelet_sync:compileDebugKotlin` and `flutter build ios --debug` all succeed.

**Not verified:** the post-fix on-device run (the Android device disconnected before I could re-run the card, so please press Verify once — expect `distinctSinks: 1` and a green verdict), and the iOS card run. The Swift test file parses but was not executed: CI has no iOS native test job and the SDK package is iOS-only, so `swift test` does not run on a macOS host.

## Not addressed here

The two amplifiers the issue also mentions, left for a separate change: `executeFallbackHttpSync` hardcoding `connectTimeout`/`readTimeout` to 15s instead of honouring `HttpConfig.httpTimeout`, and `syncBatchBlocking` using `runBlocking` inside a `Dispatchers.IO` coroutine, which costs two threads per in-flight sync.


GalacticTitan:
## Added: the pace never leaving MOVING on Android

Pushed `9cff9dd0`, a second, independent fix found while investigating a report that the example app's pace stayed MOVING on Android. It is unrelated to the sync sink but lands in the same 3.7.2 release train, so it rides along here rather than in a competing PR against the same CHANGELOG section. Happy to split it out if you'd rather review them separately.

### What the device log showed

The state machine does work — it went stationary 30 s after start:

```
15:29:33.779  State transition: MOVING -> SLOWING
15:30:03.832  State transition: SLOWING -> STATIONARY
15:30:03.860  SmartMotionCoordinator: onSpeedStateChange -> isMoving=false, action=SWITCH_TO_STATIONARY_PERIODIC
```

Then the phone was picked up (`SHAKE DETECTED! mag=2.612 > threshold=2.0`), which correctly returned it to MOVING, and the next countdown never completed:

```
15:30:36.058  State transition: MOVING -> SLOWING
15:30:58.850  State transition: SLOWING -> MOVING     ← single fix at 1.56 m/s, 21.7s into a 30s window
15:31:01.741  State transition: MOVING -> SLOWING     ← countdown restarts from zero
```

Every surrounding fix reported `0.00 m/s` and the accelerometer had already logged `[STILLNESS] ★★★ sustained stillness detected (25 samples)`. One noisy GPS sample was discarding the whole `speedStationaryDelay` window, so the pace could keep restarting indefinitely. In smart mode the Rust coordinator ORs the accel and speed inputs, so the stale speed flag alone was enough to hold the device in MOVING — which is why only a manual `changePace()` (it sets *both* flags) appeared to work.

### The fix

`SpeedMotionManager.handleSlowing`/`onLocationSlowing` now require `SPEED_ABORT_FIX_COUNT` (3) consecutive above-threshold fixes before abandoning the countdown, and a low fix resets the streak so blips cannot accumulate. The countdown keeps its original start time across a blip, so a still device reaches STATIONARY on schedule.

This mirrors `MotionDetector.MOTION_ABORT_COUNT`, which already applies the same sustained-motion rule to accelerometer noise, with the same rationale in its doc comment. It is safe by construction: SLOWING is still *continuous* tracking, so confirming real motion over a couple of fixes costs no location fidelity.

Applied to both platforms — the Android and iOS state machines mirror each other, and iOS had the identical single-sample abort even though its `CLLocation.speed` is cleaner in practice.

### Tests

- `SpeedMotionManagerTest` (Robolectric): a single blip 20 s into a 30 s window leaves the state in SLOWING and still reaches STATIONARY ~10 s later; two blips are tolerated; a low fix resets the streak. All three fail without the fix (verified by reverting the source and re-running) and the suite passes with it — 13 tests, 0 failures.
- `SpeedMotionManagerTests.swift`: the same four cases mirrored, including one asserting the countdown keeps its start time across a blip.
- The pre-existing "SLOWING returns to MOVING" expectations on both platforms now feed sustained motion, which is the intended behaviour change.

### Not changed

The example app sets `shakeThreshold: 2` (below the Android default of 2.5), so simply handling the phone re-declares MOVING. That is the accelerometer doing its job on a device that genuinely moved, and it is example configuration rather than SDK behaviour — worth knowing when testing, since picking the phone up to look at the screen restarts the cycle.

GalacticTitan:
## Added: smart-motion state ownership + platform-tuned sensor defaults

Pushed `95c3485b`. This is the follow-up to the audit above: three defects in the smart-motion pipeline and the config wiring, plus the answer to "what reintroduced this".

### Nothing reintroduced it — it never worked in smart mode

The single-fix SLOWING abort is original code from `5421e7a0 feat: Speed-Based Motion Detection (#83)` (2026-05-22); the first version of that function is identical to the one just replaced. What changed is which machine decides the pace: `41294020 chore: complete phase 3 - rust port of smart motion and battery budget engine` (2026-05-25) flipped the example app from `MotionDetectionMode.accelerometer` to `.smart`.

In accelerometer mode the pace flips purely on `MotionDetector` — 25 still samples then `stopTimeout` — and GPS speed never enters the decision, which is why it used to change in about a minute. In smart mode the Rust coordinator ORs both inputs, so the accelerometer's timeout can't flip the pace while a noisy speed flag says moving. Note `daf61dc8 fix(motion): keep stillness sampler alive and require sustained motion to abort stop-timeout` fixed exactly this class of bug on the accelerometer side (`MOTION_ABORT_COUNT`); the speed side never got the same treatment.

### Defects fixed

**1. The accelerometer input was inert after a start that began in MOVING.** The Rust coordinator initialises `is_accel_moving = false` and `on_accel_state_change()` early-returns on an unchanged flag, and `start()` never seeded it. So the stop-timeout would fire, report stationary, and the coordinator would see no transition and emit nothing — leaving the speed machine as the only input able to change the pace. `start()` now seeds the flag from the state it starts in, on both platforms.

**2. Android only synced the coordinator's tracking mode in `initialize()`, from the persisted mode.** A session that ended stationary left the coordinator believing it was already in `StationaryPeriodic`, and `evaluate_state()` only emits a stationary action while it believes the mode is `Continuous` — so the pace could never leave MOVING again until the process was killed. `start()` now re-syncs. (This is also why forcing the pace worked while automatic transitions didn't: `changePace()` sets *both* flags.)

**3. `MotionDetector` wrote `state.isMoving` itself even in smart mode**, where it is one of two inputs and the coordinator owns the decision. That left `getState().isMoving` disagreeing with the last `onMotionChange` event, and with actual GPS behaviour, whenever the coordinator stayed continuous.

### Config wiring audit

**Correct, no change needed.** `speedStationaryDelay` — the 30 s timer — is wired end to end: Dart default 180 → `toMap()` → typed Pigeon field → `ConfigManager.getSpeedStationaryDelay()` (default 180 on both platforms) → cached in `SpeedMotionManager.start()` with a `>= 0` clamp and an explicit warning at 0. Because it is cached at start, a mid-session change would be ignored — but both `setConfig` implementations already list every motion key as a restart key (the #230 fix), so the pipeline restarts and re-caches. `stopTimeout` is read live at countdown start. Android's motion defaults match Dart's 1:1.

**Broken, fixed here.** `shakeThreshold`, `stillThreshold` and `stillSampleCount` are a single cross-platform scalar, and Dart's defaults are the *Android* values — sent unconditionally by any app that configured any motion field. iOS converts m/s² to g, so `stillThreshold: 0.4` arrived as `0.04 g`, roughly four times stricter than the `0.15 g` iOS default it was supposed to keep, and `stillSampleCount: 25` dwelt ~2.5 s at iOS's 10 Hz instead of the intended ~5 s. They are now only transmitted when the app sets them: the Pigeon fields are nullable, both `TraceletHostApiImpl`s omit the key, and each platform falls back to its own tuned value.

Not a breaking change: reading `MotionConfig.shakeThreshold` still reports `2.5`, and setting a value — even one equal to the default — is honoured, carried by the new `hasExplicit*` flags. Also corrected iOS's `getStillSampleCount()` fallback from 30 to 50 (≈5 s at 10 Hz), and deleted three `private static let` constants in the iOS `MotionDetector` that documented the tuned values but were never referenced, so the documented defaults silently disagreed with the effective ones.

### Tests

- `SmartMotionCoordinatorAccelSeedTest` — an unseeded flag makes a stationary accel report a no-op; seeding lets the accelerometer complete the switch once speed agrees (using a walking-speed fix so the hand-tremor override doesn't short-circuit it); a seeded accel alone cannot force stationary while speed still moves.
- `MotionDetectorSmartModeStateOwnershipTest` — smart mode reports the transition but leaves `isMoving` alone; accelerometer mode still owns it. Verified to fail without the fix.
- `motion_threshold_defaults_test.dart` — 10 cases: omission when unset, transmission when explicit, explicit-equal-to-default, and `fromMap` round-trips that must not promote a platform default into an override.

`sdk/android`: 289 tests, only the 2 pre-existing `GeofenceManagerExitAccuracyTest` failures, which fail on this branch with these changes stashed — they've rotted because CI doesn't run the `sdk/android` suite (it only runs `:tracelet_android:testDebugUnitTest`). Worth a separate issue.

Also green: `packages/tracelet` 476 + 10 tests, `packages/tracelet_platform_interface` 303, Pigeon regenerated from the schema (not hand-edited), `melos run format` / `melos run analyze`, `:app:compileDebugKotlin`, `:tracelet_android:compileDebugKotlin`, `flutter build ios --debug`.

**Not verified on-device** — no Android device attached at the moment. Worth one run of Start → sit still → expect STATIONARY in ~30 s, and a second Start after ending stationary to confirm defect 2 is gone.

GalacticTitan:
## Added: issue #288 + its verification card

Filed [#288](https://github.com/Ikolvi/Tracelet/issues/288) for the smart-motion bug (all three defects, with the log evidence), linked the 3.7.2 CHANGELOG entries to it, and added the example card in `1ed63a61` so the fix can be checked on a device instead of argued about — same convention as the #286 card in this PR.

### How the card verifies it

It observes the real pipeline rather than poking at internals: subscribe to `onSpeedMotionChange` and `onMotionChange`, print a timestamped transition timeline, and time how long a still device takes to reach STATIONARY against a `speedStationaryDelay + 20s` budget. That is the user-visible symptom, and the timeline separates the three defects on its own:

| observation | verdict |
| --- | --- |
| STATIONARY reached, no `SLOWING → MOVING` flips | 🟢 fixed |
| STATIONARY reached but flips on the way | 🟠 passed late — device really moved, or sustained noise |
| flips and never settles | 🔴 defect 1 — a single noisy fix still discards the countdown |
| speed machine hits STATIONARY, pace never changes | 🔴 defect 2/3 — coordinator swallowed it (unseeded accel flag, or stale tracking mode) |
| never enters SLOWING | ⚠️ inconclusive — no fix below `speedMovingThreshold` |

It refuses to guess: if tracking is stopped, or the pace is already STATIONARY, it says what to do instead of reporting a meaningless pass. It also warns that the example sets `shakeThreshold: 2` (below Android's 2.5 default), so touching the phone legitimately restarts the countdown — which is exactly what happened at `15:30:35` in the log that started this thread.

### To check it

Run the example in debug, press **Start**, wait for the pace to read MOVING, put the phone on a desk, then Recent Issues → search `288` → **Verify** and leave it alone for ~50 s. Expect 🟢 with STATIONARY around the 30 s mark and zero flips.

For defect 3 specifically: end a session while stationary, then Start again and re-run — before the fix the coordinator stayed convinced it was already in `StationaryPeriodic`, and the pace could never leave MOVING.

### PR state

Four commits, all in the 3.7.2 train:

- `cc160433` process-wide `TraceletSyncSink` (#286)
- `9cff9dd0` GPS speed noise no longer restarts the SLOWING countdown (#288)
- `95c3485b` smart-mode state ownership + platform-tuned sensor defaults (#288)
- `1ed63a61` #288 card + CHANGELOG links

`melos run format` and `melos run analyze` clean; `packages/tracelet` 486 tests, `packages/tracelet_platform_interface` 303, `sdk/android` 289 (only the 2 pre-existing `GeofenceManagerExitAccuracyTest` failures, which also fail with these changes stashed); Pigeon regenerated from the schema rather than hand-edited; `flutter build apk --debug`, `:tracelet_android:compileDebugKotlin` and `flutter build ios --debug` all pass. Still unverified on-device — no Android device attached here.

GalacticTitan:
## Coverage audit: every fix in this PR now has a changelog entry and a card

Pushed `2ddf72dc`. Auditing what this release actually changed against what it documents and verifies turned up two changelog gaps and one verification gap.

### Changelog gaps

- **`packages/tracelet_sync`** said only "Version alignment with tracelet 3.7.2" — even though the #286 fix *lives in that package* (both native plugins). Anyone reading the changelog of the package they depend on would have seen nothing about the OOM fix. Entry added.
- **`packages/tracelet_platform_interface`** likewise, and its Pigeon schema changed: `TlMotionConfig.shakeThreshold` / `stillThreshold` / `stillSampleCount` are now nullable so "unset" can cross the channel. Entry added.

Distribution now, counting `**FIX**` bullets under `## 3.7.2`:

| package | entries | why |
| --- | --- | --- |
| `tracelet`, `tracelet_android`, `tracelet_ios`, `sdk/android`, `sdk/ios` | 6 | all four fixes plus the two #286 halves |
| `tracelet_sync` | 1 | the #286 sink fix, which is its own code |
| `tracelet_platform_interface` | 1 | the nullable Pigeon fields |
| `tracelet_web`, `tracelet_doctor`, `tracelet_supabase`, `tracelet_firebase` | 0 | untouched — version alignment only |

### Verification gap

The #288 card timed the pace transition but was silent about two of the fixes in `95c3485b`. It now also checks:

- **`getState()` must agree with the last pace event.** `MotionDetector` used to write `isMoving` itself in smart mode, where the coordinator owns it, so the reported state could disagree with the event and with actual GPS behaviour. A mismatch is now its own red verdict rather than being invisible.
- **The thresholds the native SDK is actually using**, via a new `debugIssue288EffectiveThresholds` handler on both platforms. This one is neat because the example sets `shakeThreshold: 2` explicitly and leaves the other two alone, so a single run checks both halves: the app-set value must be honoured, and the unset ones must show the platform's own tuned default (`0.4 m/s²` / 25 on Android, `0.15 g` / 50 on iOS). If an unset threshold is not the platform default, the card names it as still being pushed down from Dart.

That closes the loop on the audit findings — previously they were only covered by Dart unit tests, which prove the payload omits the keys but say nothing about what the device ends up using.

### Verified

`flutter analyze` clean, `melos run format` / `melos run analyze` clean, `flutter build ios --debug` and `:app:compileDebugKotlin` pass. The on-device run of both cards is still outstanding — no Android device attached here.

Issue #318, opened 2026-08-07
diagnostics: killed-state failures leave no usable evidence — the diagnostics are debug-level and dropped by default


GalacticTitan:
## Summary

The diagnostics needed to explain a background or killed-state failure are logged at `debug`, which the default log level discards. The one class of problem developers cannot reproduce on demand — behaviour with no UI attached and no logcat connected — is therefore also the class that leaves no usable evidence behind.

## Detail

The SDK already logs a great deal of useful killed-state detail. Almost all of it is `debug`:

```kotlin
TraceletLog.debug("Bootstrapping native tracking after boot/task-removal (trackingMode=…)")
TraceletLog.debug("Boot SMART: MotionDetector state changed — isMoving=$isMoving")
TraceletLog.debug("Sticky restart detected — bootstrapping native tracking")
```

`TraceletLogger.logInternal()` drops anything above the configured level before both the logcat write and the `insertLog` persist:

```kotlin
val configLevel = config.getLogLevel()
if (configLevel == LEVEL_OFF || level > configLevel) return
```

Defaults differ by host, and both leave the evidence out:

- **Flutter** — `LoggerConfig.logLevel` defaults to `LogLevel.info`, so the log table is populated, but every line above is `debug` and therefore dropped. The table has content and none of it answers the question.
- **Direct Android SDK consumers** — `ConfigManager.DEFAULT_LOG_LEVEL = 0 // OFF`, so nothing is persisted at all.

A third gap compounds it: `TraceletLog` falls back to raw logcat until the SDK attaches a logger, and the logger cannot persist until its database is open. On a cold boot `LocationService` runs before both, so entries describing a bootstrap that *never completed* (see #317) are exactly the ones lost.

The practical consequence: a user reports "tracking stopped updating while my phone was idle overnight", and there is nothing to look at. Asking them to raise `logLevel` and reproduce only works for bugs that reproduce on demand — which these, by their nature, do not.

## What is *not* the problem

**Retention is already implemented and bounded.** `pruneOldLogs()` enforces a row cap derived from the level (500 / 1000 / 2000) *and* `logMaxDays` (default 3), so log growth is capped on two axes. This is not a request to add retention.

## Proposal

A small, curated, **always-on** lifecycle channel that bypasses the level gate:

- motion-state transitions, on both the in-app and killed-state paths, so the two can be compared in one trace;
- service start / stop / sticky restart, and the decision taken on each;
[...]

PR #336, opened 2026-08-10
fix(motion): stop filtered fixes reporting 0 m/s and stranding a drive in STATIONARY

### LINKED ISSUE #332: Filtered locations report speed 0.0, so the speed motion machine declares STATIONARY mid-drive
## Summary

While driving, tracking silently downgrades from continuous to STATIONARY_PERIODIC. Shaking the phone by hand brings it back. Reproduced on iOS 26.5.2 with `motionDetectionMode: smart`; the same defect exists on Android because the root cause is in the shared Rust core.

## Root cause

`LocationProcessorResult::filtered()` and `::error()` hardcode `effective_speed: 0.0`:

```rust
// sdk/rust-core/core/src/algorithms/location_processor.rs
fn filtered(reason: &str) -> Self {
    Self {
        accepted: false,
        effective_speed: 0.0,   // <-- always zero, whatever the device was doing
        ...
    }
}
```

Both hosts deliberately feed **every** fix — accepted or not — into the GPS-speed motion state machine, and they feed it `result.effectiveSpeed`:

```swift
// sdk/ios/Sources/TraceletSDK/location/LocationEngine.swift:1218
// IMPORTANT: Always feed the speed motion state machine, even when the
// Rust processor rejects the location ...
speedSink?(result.effectiveSpeed)
```
```kotlin
// sdk/android/.../location/LocationEngine.kt:1147
speedMotionSpeedSink?.invoke(result.effectiveSpeed)
```

So every distance-filtered fix injects a fabricated **0.00 m/s** sample into `SpeedMotionManager`.

This is not a rare event — it is the *majority* of fixes while driving. With `auto-tune: 'vehicle' -> distanceFilter=30.0m` and ~1 Hz fixes at ~10 m/s, roughly two of every three fixes are `DISTANCE_FILTER`-rejected. The state machine therefore sees a stream dominated by zeros while the car is on a highway.

## Failure mechanism, from a real bug report

Device doing 8-11 m/s throughout. Odometer confirms 28.8 km travelled.

1. A filtered fix injects 0.00 and trips the machine out of MOVING:
```
14:06:45 [SpeedMotion] MOVING -> SLOWING (speed=0.00 < threshold=1.50)
14:06:45 [Tracelet] Location filtered by Rust processor: DISTANCE_FILTER
```
   The immediately preceding accepted fix was `speed=8.750787785460494`.

2. `SpeedMotionManager.handleSlowing` needs `speedAbortFixCount = 3` **consecutive** above-threshold fixes to return to MOVING. The interleaved fabricated zeros hit the low-speed branch and reset `highSpeedCount = 0` every time, so the counter never advances past 1 — visible as the same `1/3` over and over while the car is doing 8-11 m/s:
```
14:07:00 [SpeedMotion] SLOWING: ignoring high-speed fix 1/3 (speed=9.73 >= threshold=1.50)
14:07:03 [SpeedMotion] SLOWING: ignoring high-speed fix 1/3 (speed=10.48 >= threshold=1.50)
14:07:06 [SpeedMotion] SLOWING: ignoring high-speed fix 1/3 (speed=10.58 >= threshold=1.50)
14:07:10 [SpeedMotion] SLOWING: ignoring high-speed fix 1/3 (speed=8.77 >= threshold=1.50)
14:07:14 [SpeedMotion] SLOWING: ignoring high-speed fix 1/3 (speed=8.53 >= threshold=1.50)
```

3. The SLOWING countdown therefore runs to completion and the machine declares STATIONARY at 8.5 m/s:
```
14:07:15 [SpeedMotion] SLOWING timer expired -> STATIONARY
```

4. Meanwhile the accelerometer independently declared stillness — a phone held still in a smoothly moving car reads near-zero, which is the exact scenario `SpeedMotionManager` exists to cover:
```
14:06:14 Accelerometer detected sustained stillness (50 samples), starting stop-timeout
14:07:14 startStopTimeoutCountdown: Timer FIRED! Transitioning to STATIONARY
```

5. `SmartMotionCoordinator` is a logical OR of the two inputs. Both are now false, so the OR collapses and tracking is downgraded mid-drive:
```
14:07:15 SmartMotionCoordinator: onSpeedStateChange -> isMoving=false, action=switchToStationaryPeriodic
14:07:15 [Tracelet] switchToStationaryPeriodic: stopping continuous, starting periodic timer
14:07:15 [Tracelet] CLBackgroundActivitySession stopped
```

The GPS-speed machine is precisely the safeguard against the accelerometer being fooled by a smooth ride — its own class comment says so:

> Designed for vehicle tracking where accelerometer-based stop detection is unreliable (phone on a smooth dashboard reads near-zero even at highway speed).

Feeding it fabricated zeros disables that safeguard exactly when it is needed.

6. Recovery only comes from shaking the device, which is what the reporter observed:
```
14:08:19 [SpeedMotion] STATIONARY -> MOVING (wakeConfirm reached)
14:08:19 SmartMotionCoordinator: Switching to CONTINUOUS
```

## Impact

Continuous tracking silently degrades to periodic sampling during a real trip, dropping the trip's location fidelity, whenever the phone is held or resting still rather than being jostled. Battery-saving machinery activates while the vehicle is moving.

## Fix

A rejected fix still carries a trustworthy speed reading; report it instead of fabricating zero. `filtered()`/`error()` should take the resolved effective speed. Mock/spoof rejections are the exception — those fixes are untrusted in full and should keep reporting 0.

## Affected

- `sdk/rust-core/core/src/algorithms/location_processor.rs` (root cause, shared)
- `sdk/ios/Sources/TraceletSDK/location/LocationEngine.swift`
- `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/location/LocationEngine.kt`

### LINKED ISSUE #333: SmartMotionCoordinator treats an unavailable GPS speed (-1) as standing still and overrides a moving accelerometer
## Summary

`SmartMotionCoordinator.onSpeedStateChange` reads the raw platform speed off the last location to decide whether to overrule a *moving* accelerometer. On both platforms the raw value is `-1` when the fix carries no speed (`CLLocation.speed` and Android's `Location.speed` both use a negative / unset sentinel), and `-1` passes the "near zero" test. An unknown speed is therefore read as proof the device is standing still, and the accelerometer — the one input that *is* reporting motion — gets overridden to false.

## Code

iOS, `sdk/ios/Sources/TraceletSDK/SmartMotionCoordinator.swift`:

```swift
let lastSpeed = sdk?.locationEngine.getLastLocation()?.speed ?? 0
if lastSpeed <= 0.15 {
    // "hand tremor" — override the accelerometer to false
    _ = coreCoordinator?.onAccelStateChange(isMoving: false)
}
```

Android, `sdk/android/.../motion/SmartMotionCoordinator.kt`:

```kotlin
val lastSpeed = locationEngine.getLastLocation()?.speed ?: 0f
if (lastSpeed <= 0.15f) {
    coreCoordinator.onAccelStateChange(false)
}
```

Three problems in those two lines:

1. **`-1` (speed unavailable) satisfies `<= 0.15`.** A fix with no speed reading is indistinguishable from a stationary one, so the coordinator overrules a moving accelerometer on missing data. Indoors, in tunnels, and on network-provider fixes this is common.
2. **It reads the raw platform speed rather than the resolved effective speed.** `LocationEngine` already maintains `lastEffectiveSpeed`, which falls back to a distance/time-derived speed exactly when the platform reading is absent. That is the value every other consumer uses.
3. **The log line is wrong above walking pace.** Any speed over the threshold prints "suggests walking", which in the field produces entries like:
```
14:06:40 SmartMotionCoordinator: GPS speed 11.26 m/s suggests walking - trusting accel, staying continuous.
```
   11.26 m/s is 40 km/h. Someone reading a bug report sees the coordinator describing a car as a pedestrian.

## Impact

The override collapses the coordinator's logical OR to the speed machine alone at the moment the accelerometer disagrees with it — the moment the OR exists to arbitrate. Combined with the fabricated zeros in the sibling issue this is a second path to "tracking went stationary while I was moving".

## Fix

Use the resolved effective speed, treat a negative/unavailable reading as "unknown" rather than "zero" (unknown must not be grounds for overruling a positive motion signal), and make the log say what the branch actually concluded.

### LINKED ISSUE #334: Speed-motion transitions and mode switches are debug-only, so a default-level bug report cannot explain a mid-trip downgrade
## Summary

A bug report from a device that downgraded to periodic tracking mid-drive contained the raw material to diagnose it, but only because that user happened to be running with `logLevel: debug`. At the shipped defaults (`info` for Flutter, `off` for a direct SDK consumer) the same report would have shown a tracking-mode downgrade with **no record of which input decided it or what speed it decided on**.

`TraceletLogger.lifecycle` (#318) already exists for exactly this: a curated, low-frequency set of events that bypasses the level gate and persists to the SQLite log store, surfaced by Doctor under `## Session lifecycle`. Its own doc comment lists "motion-state transitions" and "tracking-mode switches" as belonging on the channel. The GPS-speed motion machine's transitions are neither — they are `debug` only.

## What the lifecycle trace showed

```
2026-08-10 14:08:28 session: stop - was mode=TraceletTrackingMode(rawValue: 0) enabled=true isMoving=true
2026-08-10 14:08:19 motion: isMoving=true mode=MotionDetectionMode(rawValue: 2) launchedInBackground=false
2026-08-10 14:07:14 motion: isMoving=false mode=MotionDetectionMode(rawValue: 2) launchedInBackground=false
```

That records the accelerometer flipping, and nothing else. Absent from the always-on trace:

- the `SpeedMotionManager` MOVING -> SLOWING -> STATIONARY transitions, and **the speed each one was decided on** — the fabricated `speed=0.00` in those lines is the entire bug;
- the coordinator's resulting tracking-mode switch and which of its two inputs was false;
- the enum values are printed as `rawValue: 0` / `rawValue: 2` rather than names, so reading the trace requires the source.

## Also missing from the report

- **`## Active configuration` rendered as `{}`.** It mirrors `Tracelet.activeConfig`, a Dart-side in-memory copy of the last `Config` passed in, so a report generated after a background relaunch shows nothing. `speedMovingThreshold`, `speedStationaryDelay` and `stopTimeout` all had to be inferred from timestamps.
- **`## Location filter (in force vs. configured)` was empty** ("_No location processor - tracking has not run in this process_") for the same reason: the process that ran the trip had already been replaced by the time Doctor ran.

## Proposal

Put the speed motion machine's state transitions and the coordinator's mode switches on the always-on lifecycle channel, with the deciding speed and the state of both inputs included. These fire a handful of times per trip, not per fix, which is the affordability bar the channel's doc comment sets.

Log the speed handed to the motion state machine alongside the filter reason on the per-fix `debug` line too, so the accepted and rejected paths can be compared directly instead of inferred.

### LINKED ISSUE #335: iOS SpeedMotionManager emits the speedmotion event twice per transition
## Summary

`SpeedMotionManager` on iOS emits the `speedmotion` event **twice** for most transitions. Android emits once. A Dart listener therefore sees duplicate `MOVING -> SLOWING`, `SLOWING -> MOVING` and `SLOWING -> STATIONARY` events on iOS only, and the state is persisted twice per transition.

## Cause

`onLocation` snapshots the state, delegates to a handler, and then emits if the state changed:

```swift
// sdk/ios/Sources/TraceletSDK/motion/SpeedMotionManager.swift
public func onLocation(speed: Double) {
    let previousState = state
    switch state {
    case .moving:     handleMoving(speed: effectiveSpeed, now: now)
    ...
    }
    if state != previousState {
        persistState()
        emitEvent(previous: previousState, current: state)   // (2)
    }
}
```

But three of the handlers already emit for themselves:

```swift
private func handleMoving(speed: Double, now: TimeInterval) {
    if speed < speedMovingThreshold {
        let previousState = state
        state = .slowing
        ...
        if state != previousState {
            persistState()
            emitEvent(previous: previousState, current: state)   // (1)
        }
    }
}
```

Both fire, with identical payloads. `handleStationary` is the odd one out — it emits nothing of its own and relies solely on the outer block, which is why `STATIONARY -> MOVING` is the only transition that emits once.

The result is an inconsistent contract: the same class emits once or twice depending on which edge was taken.

## Why it matters now

Beyond the duplicate Dart events, this blocks putting the machine's transitions on the always-on lifecycle channel (#334): a doubled trace entry would misrepresent a single transition as two, in exactly the diagnostic output meant to settle "what happened during that trip".

## Fix

Give iOS the single `transitionTo` choke point Android already has (`SpeedMotionManager.kt:345`), which persists, emits and logs exactly once, and have the handlers call it instead of hand-rolling the sequence.

## Affected

- `sdk/ios/Sources/TraceletSDK/motion/SpeedMotionManager.swift`

### LINKED ISSUE #337: Android SpeedMotionManager emits a speedmotion event and a trace line for a no-op transition
## Summary

Android's `SpeedMotionManager.transitionTo` emits a `speedmotion` event and writes a lifecycle entry even when the "transition" does not change the state. Calling `changePace(false)` on a machine that is already STATIONARY produces an event whose `previousState` equals its `state`, and a trace line reading `speed-motion: STATIONARY -> STATIONARY`.

Found by the #335 verification card on a real device:

```
✅ #335 a forced stop emits one event, not two — changePace(false) produced
   exactly one stationary → stationary event
❌ no event reports a transition to the state it came from — REGRESSED — 1
   event(s) with previousState == state
```

## Cause

```kotlin
// sdk/android/.../motion/SpeedMotionManager.kt
private fun transitionTo(newState: SpeedMotionState, reason: String) {
    val previousState = currentState
    currentState = newState
    ...
    events.sendSpeedMotionChange(eventData)          // fires unconditionally
    TraceletLog.lifecycle("speed-motion: ${previousState.name} -> ${newState.name} — $reason")
}
```

There is no `newState == previousState` guard. iOS's equivalent choke point has one:

```swift
// sdk/ios/Sources/TraceletSDK/motion/SpeedMotionManager.swift
private func commitTransition(from previousState: SpeedMotionState, because reason: String) {
    guard state != previousState else { return }
    ...
}
```

`onManualPaceChange` calls `transitionTo` unconditionally on both branches, so any `changePace()` that agrees with the current state produces one of these.

**iOS is not affected.** Every one of its five commit sites was checked: `handleMoving` only runs while MOVING and sets SLOWING; `handleSlowing` sets MOVING or STATIONARY from SLOWING; the countdown work item is behind `if self.state == .slowing`; the wake path sets MOVING from STATIONARY. All four are real edges by construction. `onManualPaceChange` is the one site that can be a no-op, and `commitTransition`'s guard absorbs it. This is a divergence to close on the Android side, not a shared defect.

## Impact

1. **A meaningless event on the public stream.** A `SpeedMotionEvent` with `previousState == state` tells a consumer nothing, and any listener that reacts to a stationary transition — switching UI, starting a timer, logging a trip segment — is driven again for a transition that never happened.

2. **It corrupts the diagnostic trace #334 just added.** `speed-motion: STATIONARY -> STATIONARY — manual pace change` is a fabricated transition in the one output meant to settle what actually happened during a trip. That trace is only worth having if every line in it is a real edge.

3. **Platform divergence.** iOS suppresses these, Android does not, so the same app behaves differently on the two platforms for the same call.

## Fix

Gate the event emission and the lifecycle entry on the state actually changing.

The persistence and the `state.isMoving` write should stay unconditional: `transitionTo` is also where `isMoving` is re-synced, and `SmartMotionCoordinator.handleAction` writes that field directly too, so a no-op call is a legitimate re-assertion of it. Only the two *observable* outputs are wrong.

Note that the `switchToStationary()` / `switchToContinuous()` callbacks in `onManualPaceChange` should likewise keep firing — re-asserting the tracking mode on an explicit `changePace()` is intended.

## Affected

- `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/motion/SpeedMotionManager.kt`

GalacticTitan:
## The report

Driving with the phone held in hand. Tracking dropped to STATIONARY mid-journey; shaking the phone brought it back. iOS 26.5.2, `motionDetectionMode: smart`. Odometer confirms 28.8 km travelled.

The root cause is in the shared Rust core, so Android had it too.

## Root cause (#332)

`LocationProcessorResult::filtered()` hardcoded `effective_speed: 0.0`. Both hosts feed **every** fix into the GPS-speed motion state machine — accepted or not, deliberately, because a parked device's fixes are all distance-filtered and the machine would otherwise never leave MOVING — and they feed it `result.effectiveSpeed`. So every rejection injected a fabricated *"stopped"* sample.

In a vehicle that is the majority of fixes. At the `vehicle` auto-tune's 30 m distance filter and ~1 Hz fixes, roughly two of every three fixes at 10 m/s are rejected.

From the reporter's log, device doing 8–11 m/s throughout:

```
14:06:45 [SpeedMotion] MOVING -> SLOWING (speed=0.00 < threshold=1.50)
14:06:45 [Tracelet] Location filtered by Rust processor: DISTANCE_FILTER
```

The accepted fix immediately before it carried `speed=8.75`.

`handleSlowing` needs `speedAbortFixCount = 3` **consecutive** above-threshold fixes to return to MOVING. The interleaved zeros hit the low-speed branch and reset `highSpeedCount = 0` every time, so the counter never advanced past 1 — the same `1/3` five times in a row while the car was on a highway:

```
14:07:00 SLOWING: ignoring high-speed fix 1/3 (speed=9.73 >= threshold=1.50)
14:07:03 SLOWING: ignoring high-speed fix 1/3 (speed=10.48 >= threshold=1.50)
14:07:06 SLOWING: ignoring high-speed fix 1/3 (speed=10.58 >= threshold=1.50)
14:07:10 SLOWING: ignoring high-speed fix 1/3 (speed=8.77 >= threshold=1.50)
14:07:14 SLOWING: ignoring high-speed fix 1/3 (speed=8.53 >= threshold=1.50)
14:07:15 [SpeedMotion] SLOWING timer expired -> STATIONARY
```

Meanwhile the accelerometer independently declared stillness — a phone held still in a smooth-riding car reads near-zero, which is the exact scenario `SpeedMotionManager` exists to cover, per its own class comment:

> Designed for vehicle tracking where accelerometer-based stop detection is unreliable (phone on a smooth dashboard reads near-zero even at highway speed).

With both inputs false the coordinator's logical OR collapsed:

```
14:07:15 SmartMotionCoordinator: onSpeedStateChange -> isMoving=false, action=switchToStationaryPeriodic
14:07:15 [Tracelet] switchToStationaryPeriodic: stopping continuous, starting periodic timer
```

### Fix

`filtered()`/`error()` take the resolved speed. Rejections that impugn the fix itself still report `0.0` — a spoofed location, and a teleport, where the position-derived speed *is* the quantity that just failed the check.

## Also fixed, found in the same path

**#333 — an unknown speed counted as standing still.** `SmartMotionCoordinator` read `getLastLocation()?.speed` to decide whether to overrule a *moving* accelerometer. `CLLocation.speed` is `-1` when the fix carries no speed, and `-1` sails through `<= 0.15`; Android's `Location.speed` reports `0.0` in the same situation. Missing data was treated as proof the device was parked, and the one input reporting motion got overridden. It now reads the resolved `lastEffectiveSpeed` — whose declaration names this hazard outright ("the cached CLLocation.speed may be stale, 0, or -1") — and treats "no speed resolved yet" as unknown, which never overrules a positive motion signal. The log line that described 11.26 m/s as "suggests walking" now names the threshold it compared against.

**#334 — the trace could not explain the downgrade.** Every `[SpeedMotion]` line is `debug`. At the shipped defaults (`info` for Flutter, `off` for a direct SDK consumer) none of them reach the store, so this was only diagnosable because the reporter happened to be running with `logLevel: debug`. `TraceletLogger.lifecycle` (#318) already exists for exactly this and its doc comment lists "motion-state transitions" and "tracking-mode switches" as belonging on it. Now on the always-on channel:

```
speed-motion: MOVING -> SLOWING — speed=0.00 < threshold=1.50
speed-motion: restored STATIONARY (lowSpeedCount=0, wakeCount=0)
smart-motion: switching to STATIONARY_PERIODIC — accelMoving=false speedMoving=false lastSpeed=8.53m/s
```

State names are spelled out rather than logged as `rawValue: 2`, since the trace is read by a human in a pasted report. The per-fix `debug` filter line now carries the speed handed to the machine, so accepted and rejected fixes can be compared directly. These fire a handful of times per trip, not per fix — the affordability bar the channel documents — and retention is unchanged.

**#335 — iOS emitted `speedmotion` twice per transition.** The inner handler persisted-and-emitted, then `onLocation` did it again; `handleStationary` was the odd one out and relied only on the outer block, which is why `STATIONARY -> MOVING` was the only edge that emitted once. Replaced with a single `commitTransition` choke point mirroring Android's `transitionTo`. A doubled entry would also have misrepresented one transition as two in the new trace.

## Tests

- **Rust (6 new, 16 in file passing):** the reject-path speed contract per filter reason, including `a_drive_never_reports_a_fabricated_zero_speed` — 30 fixes at 10 m/s through a 30 m filter, asserting every result reports 10.0 and that fixes are genuinely being rejected so the scenario is not vacuous.
- **Swift (8 new, 82 passing):** `SpeedMotionManagerTests.swift` was written but never listed in `Package.swift`'s test `sources`, so **none of it had ever run**. Wired in, its delegate double brought up to date against the current `SpeedMotionDelegate`, plus one-event-per-transition coverage and the drive scenario.
- **Kotlin (9 new, 335 passing):** the coordinator's unknown/near-zero/vehicle-speed branches and the drive scenario. Two existing tests asserted the old behaviour — overriding a moving accelerometer with no GPS fix at all — and were migrated to assert what they said they meant.

`melos format` and `melos analyze` clean.

## Verification cards

`example/lib/issues/issue_33{2,3,4,5}_card.dart`, registered newest-first in the issues tab. #334 and #335 verify their contracts outright (#334 raises `logLevel` to `error` first, so anything it then finds reached the store despite the gate). #332 and #333 observe the live consequence and report **inconclusive** rather than green when the device never moved — a stationary device legitimately goes STATIONARY, and calling that a pass would hide the bug.

## Found by running the card (#337)

The #335 card, run on a device, reported its count rows green and its degenerate-edge row red:

```
✅ #335 a forced stop emits one event, not two — changePace(false) produced
   exactly one stationary → stationary event
❌ no event reports a transition to the state it came from — REGRESSED — 1
   event(s) with previousState == state
```

Android's `transitionTo` had no "did the state actually change" guard, so a `changePace()` agreeing with the current state emitted a phantom event and wrote a fabricated `speed-motion: STATIONARY -> STATIONARY` line into the trace this PR adds. iOS's `commitTransition` has always guarded it — verified across all five commit sites, and the four new Swift tests pass with no iOS source change.

`state.isMoving` stays outside the guard: `transitionTo` is where that field is re-synced and `SmartMotionCoordinator.handleAction` writes it directly too, so a no-op call is a legitimate re-assertion. The mode-switch callbacks are likewise unaffected.

The card now seeds a known state with `changePace(true)` first — without it, "one event" could not be told apart from "one no-op event", which is how this got past the first pass — and asserts the repeat call emits nothing.

Fixes #337

## Known gaps, not addressed here

`## Active configuration` in a bug report renders as `{}` because it mirrors `Tracelet.activeConfig`, a Dart-side in-memory copy of the last `Config` passed in; a report generated after a background relaunch shows nothing. The same restart empties `## Location filter (in force vs. configured)`. Both are process-scoped by construction. Noted on the #334 card.

Fixes #332
Fixes #333
Fixes #334
Fixes #335

### DIFF
--- a/example/lib/issues/issue_332_card.dart
+++ b/example/lib/issues/issue_332_card.dart
@@ -1,0 +1,221 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+
+/// Issue #332 — a location the Rust processor rejected reported
+/// `effectiveSpeed: 0.0`, so the GPS-speed motion machine was told the device
+/// had stopped once per rejected fix.
+///
+/// Both hosts feed *every* fix into `SpeedMotionManager`, accepted or not —
+/// deliberately, because a parked device's fixes are all distance-filtered and
+/// the machine would otherwise never leave MOVING. They feed it
+/// `result.effectiveSpeed`, which `LocationProcessorResult::filtered()`
+/// hardcoded to zero.
+///
+/// In a vehicle that is the common case, not a corner case: at the `vehicle`
+/// auto-tune's 30 m distance filter and ~1 Hz fixes, most of a 10 m/s drive is
+/// rejected. The machine saw a stream of zeros on a motorway, dropped
+/// MOVING → SLOWING, could never collect the three *consecutive* above-threshold
+/// fixes needed to abort (each fabricated zero reset the streak), and ran its
+/// countdown out to STATIONARY. The accelerometer had independently declared
+/// stillness — a phone held still in a smooth-riding car reads near-zero, which
+/// is the exact scenario the speed machine exists to cover — so the
+/// coordinator's logical OR collapsed and tracking downgraded to periodic
+/// mid-drive. Shaking the phone was the only way back.
+///
+/// **What this card can prove, and what it cannot.** The processor is not
+/// exposed to Dart, so the fix itself is pinned by Rust unit tests
+/// (`a_drive_never_reports_a_fabricated_zero_speed`). What this card does is
+/// observe the live consequence: it watches real fixes and the speed machine
+/// side by side, and fails if the machine leaves MOVING while the device is
+/// demonstrably above the speed threshold. That is only exercised while
+/// actually moving, so a stationary run reports **inconclusive** rather than
+/// green — see the run output.
+class Issue332Card extends StatefulWidget {
+  const Issue332Card({super.key});
+
+  @override
+  State<Issue332Card> createState() => _Issue332CardState();
+}
+
+class _Issue332CardState extends State<Issue332Card> {
+  static const _observeWindow = Duration(seconds: 45);
+
+  /// The SDK default for `speedMovingThreshold` (m/s). Fixes at or above this
+  /// are what the machine must treat as motion.
+  static const _movingThreshold = 1.5;
+
+  String _status = 'Idle';
+  bool _running = false;
+
+  void _set(String s) {
+    if (mounted) setState(() => _status = s);
+  }
+
+  Future<void> _run() async {
+    setState(() => _running = true);
+    final results = <String>[];
+    var allPass = true;
+
+    void check(String name, bool pass, String detail) {
+      results.add('${pass ? '✅' : '❌'} $name — $detail');
+      if (!pass) allPass = false;
+    }
+
+    StreamSubscription<Location>? locSub;
+    StreamSubscription<SpeedMotionEvent>? smSub;
+
+    try {
+      final speeds = <double>[];
+      final transitions = <SpeedMotionEvent>[];
+      // Speed at the moment each transition landed, so a downgrade can be
+      // attributed to the reading that caused it rather than to the run as a
+      // whole.
+      final speedAtTransition = <double>[];
+
+      locSub = Tracelet.onLocation((l) => speeds.add(l.coords.speed));
+      smSub = Tracelet.onSpeedMotionChange((e) {
+        transitions.add(e);
+        speedAtTransition.add(speeds.isEmpty ? -1 : speeds.last);
+      });
+
+      await Tracelet.start();
+      _set(
+        'Observing for ${_observeWindow.inSeconds}s…\n\n'
+        'To exercise the regression, run this while travelling with the phone '
+        'held still or resting on a seat. Standing still can only confirm the '
+        'plumbing.',
+      );
+      await Future<void>.delayed(_observeWindow);
+
+      final moving = speeds.where((s) => s >= _movingThreshold).length;
+      final maxSpeed = speeds.isEmpty
+          ? 0.0
+          : speeds.reduce((a, b) => a > b ? a : b);
+
+      // Row 1: did anything arrive at all? Without this a silent location
+      // pipeline would make every row below vacuously green.
+      check(
+        'fixes were delivered',
+        speeds.isNotEmpty,
+        speeds.isNotEmpty
+            ? '${speeds.length} fix(es), peak ${maxSpeed.toStringAsFixed(2)} m/s'
+            : 'no fixes in ${_observeWindow.inSeconds}s — check location '
+                  'permission and that GPS has a sky view. Nothing below can be '
+                  'concluded.',
+      );
+
+      // Row 2: the actual regression. A downgrade out of MOVING is only a bug
+      // when the device was above the threshold at the time.
+      final downgrades = <String>[];
+      for (var i = 0; i < transitions.length; i++) {
+        final t = transitions[i];
+        final at = speedAtTransition[i];
+        if (t.state != SpeedMotionState.moving && at >= _movingThreshold) {
+          downgrades.add(
+            '${t.previousState.name} → ${t.state.name} at '
+            '${at.toStringAsFixed(2)} m/s',
+          );
+        }
+      }
+
+      if (moving == 0) {
+        results.add(
+          '⏭️ #332 the speed machine holds MOVING while the device moves — '
+          'INCONCLUSIVE. No fix reached the '
+          '${_movingThreshold.toStringAsFixed(1)} m/s threshold in this run '
+          '(peak ${maxSpeed.toStringAsFixed(2)} m/s), so the regression was '
+          'never given a chance to fire. Re-run while travelling. This is '
+          'reported as a skip, not a pass: a stationary device legitimately '
+          'goes STATIONARY, and calling that green would hide the bug.',
+        );
+      } else {
+        check(
+          '#332 the speed machine holds MOVING while the device moves',
+          downgrades.isEmpty,
+          downgrades.isEmpty
+              ? '$moving of ${speeds.length} fixes were at or above '
+                    '${_movingThreshold.toStringAsFixed(1)} m/s (peak '
+                    '${maxSpeed.toStringAsFixed(2)}) and the machine never left '
+                    'MOVING on one of them'
+              : 'REGRESSED — left MOVING while above the threshold: '
+                    '${downgrades.join('; ')}. A rejected fix is reporting a '
+                    'fabricated speed again.',
+        );
+      }
+
+      // Row 3: the diagnostic that makes the next report readable without a
+      // debug build (#334). Cheap to assert here and it shares this bug's
+      // subject, so a regression in either shows up on the same card.
+      final logs = await Tracelet.getLogs(300);
+      final speedMotionTrace = logs
+          .where((e) => e.level.toUpperCase() == 'LIFECYCLE')
+          .where((e) => e.message.contains('speed-motion:'))
+          .toList();
+      check(
+        'the speed machine leaves an always-on trace',
+        speedMotionTrace.isNotEmpty,
+        speedMotionTrace.isNotEmpty
+            ? '${speedMotionTrace.length} `speed-motion:` lifecycle entr'
+                  '${speedMotionTrace.length == 1 ? 'y' : 'ies'}, each carrying '
+                  'the speed it decided on — e.g. "'
+                  '${speedMotionTrace.last.message}"'
+            : 'no `speed-motion:` lifecycle entries. The machine records one on '
+                  'start and one per transition, so a run that stayed in a '
+                  'single state and was already started may legitimately show '
+                  'none; a run that changed state and shows none is a #334 '
+                  'regression.',
+      );
+
+      await Tracelet.stop();
+
+      final header = allPass
+          ? '✅ SUCCESS: no fix caused the speed machine to declare a stop '
+                'while the device was moving.'
+          : '❌ FAILED — see the failing rows below.';
+
+      _set(
+        '$header\n\n${results.join('\n')}\n\n'
+        'Mechanism: LocationProcessorResult::filtered() returned '
+        'effective_speed: 0.0 for every rejected fix, and both hosts feed that '
+        'value into SpeedMotionManager on every fix by design. A 30 m distance '
+        'filter rejects most of a 10 m/s drive, so the machine was told '
+        '"stopped" two times out of three on a motorway.\n\n'
+        'Why the accelerometer did not save it: a phone held still in a '
+        'smooth-riding car reads near-zero, which is precisely why the '
+        'GPS-speed machine exists. Once it was fooled too, the coordinator OR '
+        'had two false inputs and downgraded to periodic tracking.\n\n'
+        'Coverage note: the processor has no Dart entry point, so the fix '
+        'itself is pinned by Rust unit tests. This card observes the live '
+        'consequence and is only conclusive while actually moving.',
+      );
+    } catch (e) {
+      _set('❌ FAILED: $e\n\n${results.join('\n')}');
+    } finally {
+      await locSub?.cancel();
+      await smSub?.cancel();
+      if (mounted) setState(() => _running = false);
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return IssueCardShell(
+      keywords:
+          'speed motion stationary moving vehicle drive filtered distance '
+          'filter effective speed rust processor slowing countdown periodic '
+          'downgrade smart coordinator 332',
+      title: '#332: a filtered fix no longer reports a fabricated 0 m/s',
+      description:
+          'Watches live fixes and the GPS-speed motion machine together and '
+          'fails if the machine declares a stop while the device is above the '
+          'moving threshold. Run it while travelling — a stationary run is '
+          'reported as inconclusive rather than green.',
+      status: _status,
+      running: _running,
+      onRun: _run,
+    );
+  }
+}
--- a/example/lib/issues/issue_333_card.dart
+++ b/example/lib/issues/issue_333_card.dart
@@ -1,0 +1,182 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+
+/// Issue #333 — `SmartMotionCoordinator` read the raw platform speed to decide
+/// whether to overrule a *moving* accelerometer, and an unavailable reading
+/// passed the "near zero" test.
+///
+/// The coordinator is a logical OR: continuous tracking is kept while either
+/// the accelerometer or the GPS-speed machine reports motion. When the speed
+/// machine says stationary but the accelerometer disagrees, one narrow override
+/// applies — a device physically still with the accelerometer picking up hand
+/// tremor, where GPS speed is genuinely ~0. It was gated on:
+///
+/// ```swift
+/// let lastSpeed = sdk?.locationEngine.getLastLocation()?.speed ?? 0
+/// if lastSpeed <= 0.15 { /* override the accelerometer to false */ }
+/// ```
+///
+/// `CLLocation.speed` is **-1** on a fix carrying no speed, and -1 sails
+/// through `<= 0.15`; Android's `Location.speed` reports 0.0 in the same
+/// situation. So "we don't know how fast this is" was read as "definitely
+/// parked", and the coordinator overruled the one input that was reporting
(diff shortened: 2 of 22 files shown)


GalacticTitan:
### Note for review: this branch carries two commits that belong to other PRs

`fix/speed-motion-stationary-mid-drive` was the branch being used for on-device testing of these motion fixes, and that testing needed two things that are unrelated to this PR:

- **#338 / PR #339** — `example/test_server.dart`, `example/pubspec.yaml`. The local sync server had stopped printing its QR code, so the device could not be pointed at it by scanning.
- **#340 / PR #341** — `TraceletIosPlugin.swift`, `TraceletSyncPlugin.swift`, `TraceletAndroidPlugin.kt`, `NativeSyncProvider.kt`. Diagnostics for a separate iOS/Android sync-body parity problem found while testing.

Both were developed on their own branches off `main` and opened as their own PRs; they were then cherry-picked here so the device build had them. They are byte-identical to the commits in #339 and #341 — no divergence.

**Suggested merge order:** #339 and #341 first, then rebase this branch on `main`. The duplicated commits drop out on their own and this PR narrows to the motion fixes.

The motion work itself — the reason this PR exists — is confined to:

```
sdk/rust-core/core/src/algorithms/location_processor.rs
sdk/ios/Sources/TraceletSDK/{SmartMotionCoordinator,location/LocationEngine,motion/SpeedMotionManager}.swift
sdk/ios/{Package.swift,Tests/TraceletSDKTests/SpeedMotionManagerTests.swift}
sdk/android/tracelet-sdk/src/main/kotlin/.../{location/LocationEngine,motion/SmartMotionCoordinator,motion/SpeedMotionManager}.kt
sdk/android/tracelet-sdk/src/test/kotlin/.../motion/{SmartMotionCoordinatorTest,SpeedMotionManagerTest}.kt
example/lib/issues/issue_33{2,3,4,5}_card.dart + recent_issues_tab.dart
```

Issue #319, opened 2026-08-07
killed-state: engine keeps tracking continuously after the motion subsystems settle back to stationary (Android; iOS safety net)


GalacticTitan:
## Summary

In the killed state the engine's tracking mode is switched **only** from a motion *transition*. The motion subsystems can settle back into stationary without emitting one, and when that happens the engine keeps running continuous GPS for the rest of the process lifetime — OS location indicator pinned on, fixes landing every couple of seconds — while the persisted state reads stationary.

Reported as: *"in the killed state the pace is not changing even when I am not moving, it continues in the moving state. I can see the location icon on the device even after 1 minute. When opening the app again it opens with stationary pace, but the location icon is still showing."*

## Evidence

From a device bug report (`stopTimeout: 1` minute, `motionDetectionMode: smart`, continuous mode):

The session lifecycle trace (#318) contains exactly one entry:

```
2026-08-07 17:36:04  motion (killed-state, smart): isMoving=true
```

Nothing follows for ~87 s. Over that window:

- `[SHAKE] sample #7901 … #11351` stream continuously — the accelerometer never stopped, and every sample reads `current_mag ≈ 0.02` against `threshold=2.0`. The device is demonstrably still.
- `Boot heartbeat fired` / `Boot heartbeat: lat=…` keep running — the killed-state pipeline is alive and healthy.
- Locations continue to be persisted every ~2 s at 3 m accuracy right up to the moment the app is reopened.
- Health check reports `Is moving | false`.

At teardown the boot detector logs:

```
cancelSignificantMotionListener() called — existing=true
```

The significant-motion trigger is only registered in the **stationary** branch, so the detector had already returned to its stationary configuration. Detector: stationary. Persisted state: stationary. Engine: still continuous.

## Mechanism

`MotionDetector.onManualPaceChange()` swaps the sensor set between the shake/significant-motion configuration and the stillness configuration directly. It never routes through `declareStationary()`, so it never fires `onMotionStateChanged`.

The killed-state wiring in `LocationService.startBootTracking()` drives the engine's mode exclusively from that callback:

```kotlin
detector.onMotionStateChanged = { isMoving ->
    bootSpeedMotionManager?.onManualPaceChange(isMoving)
    val action = coordinator.onAccelStateChange(isMoving)
    ...
}
```

So a settle-back that produces no transition leaves the engine where it was.

`startBootTracking()` already reconciles this — but only once, at bootstrap:

```kotlin
[...]

PR #345, opened 2026-08-10
fix(motion): a stationary start no longer deafens the SMART motion coordinator

### LINKED ISSUE #344: [Bug]: SMART motion — a start() with isMoving=false permanently deafens the accelerometer; device never leaves stationary and never syncs
## Summary

In `smart` motion detection mode, a `start()` whose committed pace is **stationary** (`isMoving == false`) permanently disables the accelerometer wake-up for the whole session. The device can be shaken continuously and `MotionDetector` reports every sample, but the SDK never enters the moving state, never starts continuous GPS, records no locations — and therefore never syncs.

Reported from a Doctor bug report where the user pressed **Start** and shook the phone for ~30 s.

## Evidence (field bug report)

```
2026-08-10 20:08:20 session: start — mode=continuous resume=false isMoving=false motionMode=MotionDetectionMode(rawValue: 2)
2026-08-10 20:08:20 speed-motion: restored STATIONARY (lowSpeedCount=2, wakeCount=0)
...
2026-08-10 20:08:48 [VERBOSE] Accelerometer detected SHAKE (magnitude: 4.29), triggering moving
2026-08-10 20:08:48 [DEBUG] handleMovingDetected: triggering isMoving=true immediately
2026-08-10 20:08:48 [LIFECYCLE] motion: isMoving=true mode=MotionDetectionMode(rawValue: 2)
2026-08-10 20:08:48 [DEBUG] SmartMotionCoordinator: onAccelStateChange -> isMoving=true, action=none   <-- swallowed
```

~130 consecutive `action=none`. Health check for the same session: `Is moving | false`, `Pending locations | 8` (all stale). No `smart-motion: switching to CONTINUOUS` line anywhere after the start.

Note this is *not* an over-sensitive shake threshold: `shakeThreshold: 2.0` m/s² ÷ 9.81 = 0.204 g, and samples of 2.02 g / 4.29 g are real motion. The detector is doing its job; the coordinator is discarding the result.

## Mechanism

`TraceletSmartMotionCoordinator.syncCurrentMode()` maps the *session* mode onto the coordinator's *posture*:

`sdk/ios/Sources/TraceletSDK/SmartMotionCoordinator.swift:28-43`
```swift
let mode: TrackingMode
switch stateManager.trackingMode {
case .continuous:
    mode = .continuous          // <-- always, regardless of the committed pace
...
coreCoordinator?.setCurrentMode(mode: mode)
```

These two are not the same thing:

* `stateManager.trackingMode` is **which start API was used**. It is set to `.continuous` by `start()` and stays `.continuous` for the whole session — `switchToStationaryPeriodicForce()` deliberately does not change it (`TraceletSdk.swift:3802`); only `stateManager.isMoving` records the pace.
* The Rust coordinator's `current_mode` is **the posture right now** — whether the engine is running continuous GPS or parked in a stationary schedule.

`start()` (`TraceletSdk.swift:465-471`) sets `isMoving = configManager.getIsMoving()` (false here), then calls `syncCurrentMode()`, which force-writes `Continuous`.

The coordinator's state machine then has both inputs stationary and `current_mode == Continuous` — a combination it emits no action for in either direction (`sdk/rust-core/core/src/state/smart_motion_coordinator.rs:99-116`):

```rust
let should_be_moving = state.is_accel_moving || state.is_speed_moving;
if should_be_moving && state.current_mode != TrackingMode::Continuous {
    ... SwitchToContinuous
} else if !should_be_moving && state.current_mode == TrackingMode::Continuous {
    ... SwitchToStationaryPeriodic
} else {
    CoordinatorAction::None            // <-- lands here forever
}
```

* Accel goes true → `should_be_moving` true, but `current_mode` is *already* `Continuous` → `None`. `switchToContinuousForce()` is never called, so `stateManager.isMoving` is never set true and `locationEngine.switchToContinuous()` never runs.
* The usual self-correction is also dead: `startSpeedMotionManager()` (`TraceletSdk.swift:3724-3732`) calls `onSpeedStateChange(false)` on a restored-stationary speed machine, which would normally emit `SwitchToStationaryPeriodic` and put `current_mode` back where a wake-up can be seen. But `SmartMotionCoordinator` lives for the whole process and `on_speed_state_change` early-returns on an unchanged flag — the previous session had already left `is_speed_moving = false`, so it dedupes to `None`.
* `reconcileTrackingMode()` (#319) cannot catch it either: it treats `stateManager.isMoving` as the authority, and `isMoving == false` matches the parked engine, so it sees no divergence.

Net effect: the coordinator is stuck in a state where the accelerometer input is inert, and stays there until the process is killed.

## Reproduction

1. `motionDetectionMode: smart`, `motion.isMoving: false` (the default).
2. Start tracking, let the speed machine settle to STATIONARY, stop tracking.
3. Start tracking again **in the same process**.
4. Walk/shake the device.

Expected: `smart-motion: switching to CONTINUOUS`, `isMoving=true`, locations recorded and synced.
Actual: `onAccelStateChange -> isMoving=true, action=none` forever; `isMoving` stays false; no locations; nothing to sync.

A first-ever start also reaches the dead state whenever the speed machine restores as stationary before the first accel event.

## Why it was not caught

`sdk/ios/Tests/TraceletSDKTests/SmartMotionCoordinatorTests.swift` exists but is not listed in the `sources:` allowlist of the test target in `sdk/ios/Package.swift`, so it has never run. Its `testOnAccelStateChangeToTrueSwitchesToContinuous` only covers `startPeriodic()` (where `trackingMode == .periodic` maps correctly) and would not have covered this case anyway.

## Android

`sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/motion/SmartMotionCoordinator.kt:114-125` has the identical mapping and `TraceletSdk.kt:950` calls it from the same place in `start()`, so the same dead state is reachable there.

GalacticTitan:
Fixes #344

## The bug

In `smart` motion mode, a `start()` whose committed pace is stationary (`isMoving == false`, the default) permanently disables motion wake-up for the rest of the process. The device can be shaken continuously — `MotionDetector` reports every sample correctly — but the SDK never enters the moving state, never starts continuous GPS, records no locations, and therefore never syncs.

From the field report:

```
20:08:20 session: start — mode=continuous resume=false isMoving=false motionMode=smart
20:08:20 speed-motion: restored STATIONARY
...
20:08:48 [VERBOSE] Accelerometer detected SHAKE (magnitude: 4.29), triggering moving
20:08:48 [LIFECYCLE] motion: isMoving=true mode=smart
20:08:48 [DEBUG] SmartMotionCoordinator: onAccelStateChange -> isMoving=true, action=none
```

~130 consecutive `action=none`, no `smart-motion: switching to CONTINUOUS` anywhere, health check `Is moving | false`.

## Root cause

`syncCurrentMode()` mapped the **session** mode onto the coordinator's **posture**. They are not the same thing:

* `stateManager.trackingMode` is which start API was called. `start()` pins it to `.continuous` for the whole session; `switchToStationaryPeriodicForce()` deliberately leaves it alone and records the pace in `stateManager.isMoving`.
* The core coordinator's `currentMode` is what the engine is doing *right now* — continuous GPS, or parked in a stationary schedule.

`start()` sets `isMoving = configManager.getIsMoving()` (false) and then calls `syncCurrentMode()`, which force-wrote `Continuous` into a coordinator whose accel and speed inputs both said stationary. `evaluate_state` emits nothing for that pair in either direction:

```rust
let should_be_moving = state.is_accel_moving || state.is_speed_moving;
if should_be_moving && state.current_mode != TrackingMode::Continuous { .. SwitchToContinuous }
else if !should_be_moving && state.current_mode == TrackingMode::Continuous { .. SwitchToStationary }
else { CoordinatorAction::None }   // <- lands here, forever
```

A shake sets `should_be_moving`, but `currentMode` is *already* `Continuous`, so it returns `None` and `switchToContinuousForce()` never runs.

Both self-corrections were closed off:

* `startSpeedMotionManager()` re-asserts a restored-stationary speed machine via `onSpeedStateChange(false)`, which would normally push `currentMode` back somewhere a wake-up is visible. But the coordinator outlives `stop()` and the core dedupes an unchanged flag — a previous session in the same process had already left `is_speed_moving = false`.
* `reconcileTrackingMode()` (#319) treats `isMoving` as the authority, and a false `isMoving` matched the parked engine, so it saw no divergence.

## The fix

Read the posture off `isMoving` rather than off the session mode. A `.continuous` session that has committed a stationary pace now syncs as `stationaryPeriodic`/`stationaryGeofences` (per `stationaryTrackingMode`), so the first real accel or speed event produces `SwitchToContinuous`. The dedicated `.geofences`/`.periodic` session modes are already postures and are unchanged.

Android carried the identical mapping (`SmartMotionCoordinator.kt`, called from the same place in `start()`) and is fixed the same way.

## Tests

* **iOS** — the mapping is extracted as a pure `coordinatorMode()` so it can be pinned directly. New `SmartMotionCoordinatorSyncModeTests` covers the mapping, drives the real Rust core through the exact field sequence (previous session left stationary → sync → shake), and keeps the **old** behaviour as a negative control so the failure mode stays documented. 8 tests.
* That suite is wired into `Package.swift`. The pre-existing `SmartMotionCoordinatorTests.swift` was never in the `sources:` allowlist, so it had never run — which is why this was not caught. (It is also stale: it asserts `isMoving == true` after a bare `start()`, which contradicts the `motion.isMoving` default. Left alone here rather than rewritten in a bugfix PR.)
* **Android** — `SmartMotionCoordinatorStationaryStartTest`, same sequence through the real core. 3 tests.
* **Example app** — `issue_344_card.dart`. Reproduces the precondition (settle stationary → stop → start again in the same process) and checks the wake-up lands. `changePace(true)` enters the coordinator through the same two inputs a physical shake uses and was swallowed identically on the broken build, so the defect is reproducible from Dart. The card states plainly that it does not cover `MotionDetector` delivering the shake — that path was never broken.

## Verification

* iOS: `xcodebuild test -scheme TraceletSDK` — 90 tests, 0 failures.
* Android: `./gradlew :tracelet-sdk:testDebugUnitTest` — BUILD SUCCESSFUL.
* `melos format` and `melos analyze` clean.

### DIFF
--- a/example/lib/issues/issue_344_card.dart
+++ b/example/lib/issues/issue_344_card.dart
@@ -1,0 +1,226 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+
+/// Issue #344 — in SMART motion mode, a `start()` whose committed pace was
+/// stationary left the coordinator permanently deaf to motion.
+///
+/// `syncCurrentMode()` mapped the *session* mode onto the coordinator's
+/// *posture*, and those are not the same thing:
+///
+/// * `stateManager.trackingMode` says **which start API was called**. `start()`
+///   pins it to `.continuous` for the whole session — the stationary switch
+///   deliberately leaves it alone and records the pace in `isMoving`.
+/// * The core coordinator's `currentMode` says **what the engine is doing right
+///   now**: continuous GPS, or parked in a stationary schedule.
+///
+/// ```swift
+/// case .continuous:
+///     mode = .continuous          // regardless of the committed pace
+/// ```
+///
+/// `start()` sets `isMoving` from `motion.isMoving` (false by default) and then
+/// syncs, so this wrote `Continuous` into a coordinator whose accelerometer and
+/// speed inputs both said stationary. The core emits no action for that pair in
+/// either direction:
+///
+/// ```rust
+/// let should_be_moving = state.is_accel_moving || state.is_speed_moving;
+/// if should_be_moving && state.current_mode != TrackingMode::Continuous { .. }
+/// else if !should_be_moving && state.current_mode == TrackingMode::Continuous { .. }
+/// else { CoordinatorAction::None }     // <- lands here, forever
+/// ```
+///
+/// A shake set `should_be_moving`, but `currentMode` was *already* `Continuous`,
+/// so the wake-up returned `None`, `switchToContinuousForce()` never ran, and
+/// `isMoving` never became true. The field report that opened this issue shows
+/// ~130 consecutive `onAccelStateChange -> isMoving=true, action=none` while the
+/// user shook the phone: no fixes recorded, and therefore nothing to sync.
+///
+/// The usual escape hatches were both closed. `startSpeedMotionManager()` re-
+/// asserts a restored-stationary speed machine, but the coordinator lives for
+/// the whole process and the core dedupes an unchanged flag, so the previous
+/// session had already left it false. And `reconcileTrackingMode()` (#319)
+/// treats `isMoving` as the authority, so a false `isMoving` matched the parked
+/// engine and looked perfectly consistent.
+///
+/// **What this card can prove.** Dart cannot inject accelerometer samples, but
+/// it does not have to: `changePace(true)` enters the coordinator through the
+/// same `onAccelStateChange` / `onSpeedStateChange` inputs the shake used, and
+/// on the broken build it is swallowed identically. The card reproduces the
+/// exact precondition — a session that settled stationary, stopped, and started
+/// again **in the same process** — and then checks the wake-up actually lands.
+///
+/// **What it does not prove.** That a *physical* shake reaches the coordinator;
+/// that is `MotionDetector`'s job and is unchanged here. The mapping itself is
+/// pinned by `SmartMotionCoordinatorSyncModeTests` (iOS) and
+/// `SmartMotionCoordinatorStationaryStartTest` (Android), both of which drive
+/// the real core through the same sequence.
+class Issue344Card extends StatefulWidget {
+  const Issue344Card({super.key});
+
+  @override
+  State<Issue344Card> createState() => _Issue344CardState();
+}
+
+class _Issue344CardState extends State<Issue344Card> {
+  String _status = 'Idle';
+  bool _running = false;
+
+  void _set(String s) {
+    if (mounted) setState(() => _status = s);
+  }
+
+  Future<void> _run() async {
+    setState(() => _running = true);
+    final results = <String>[];
+    var allPass = true;
+
+    void check(String name, bool pass, String detail) {
+      results.add('${pass ? '✅' : '❌'} $name — $detail');
+      if (!pass) allPass = false;
+    }
+
+    StreamSubscription<Location>? sub;
+
+    try {
+      // SMART is the mode the coordinator arbitrates; isMoving:false is the
+      // default and the value that triggered the bug.
+      await Tracelet.setConfig(
+        const Config(
+          motion: MotionConfig(
+            motionDetectionMode: MotionDetectionMode.smart,
+            isMoving: false,
+          ),
+        ),
+      );
+
+      final motionEvents = <bool>[];
+      sub = Tracelet.onMotionChange((loc) => motionEvents.add(loc.isMoving));
+
+      // --- Session 1: settle stationary, then stop. -----------------------
+      // This is what the field report had: a session that ended with both
+      // coordinator inputs reading stationary. The coordinator outlives stop(),
+      // so session 2 inherits those flags — which is precisely why session 2's
+      // own "re-assert stationary" call dedupes to a no-op and cannot repair
+      // the posture.
+      await Tracelet.start();
+      await Future<void>.delayed(const Duration(seconds: 2));
+      await Tracelet.changePace(false);
+      await Future<void>.delayed(const Duration(seconds: 2));
+      await Tracelet.stop();
+      await Future<void>.delayed(const Duration(seconds: 1));
+
+      // --- Session 2: the start whose committed pace is stationary. -------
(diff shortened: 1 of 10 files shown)


GalacticTitan:
Follow-up pushed: the stale `SmartMotionCoordinatorTests.swift` is now fixed and wired into `Package.swift` rather than left dead.

It turned out not to compile at all — `coordinator` was typed as the Rust `SmartMotionCoordinator` while `sdk.smartMotionCoordinator` is the native `TraceletSmartMotionCoordinator` wrapper — and on top of that it passed `"motionDetectionMode": "smart"` as a **string** to a getter that only reads `as? Int`, so the config fell back to `.activity`: even had it built, nothing in the file would have exercised smart mode. A guard test now pins the config so it cannot regress to testing nothing.

Tracking can't be started in this target — `locationEngine.start()` reaches `CLLocationManager.setAllowsBackgroundLocationUpdates`, which throws `NSInternalInconsistencyException` unless the running bundle declares the `location` background mode, and a SwiftPM xctest bundle has no app host to borrow one from. So the tests set `stateManager` directly and leave `enabled` false; the force-switch handlers bail on their own guard, the engine is never touched, and the coordinator's decision is what gets asserted. Each test also builds its own coordinator, since the SDK's survives `reset()` and was leaking motion flags between tests.

**Verified against a deliberately reverted mapping**: `testAccelWakesAContinuousSessionThatStartedStationary` fails with `"none" is not equal to "switchToContinuous"`, along with 4 of the `SmartMotionCoordinatorSyncModeTests`. These are not vacuous.

Two supporting production fixes fell out of it:

- `getState()` documents that it returns a default disabled state before `ready()`, but force-unwrapped nil managers and trapped — and `reset()` routes its own not-ready path straight into it. It returns `StateManager.disabledStateMap()` now.
- `onSpeedStateChange()` returned `Void` while its accelerometer counterpart returned the action, leaving the speed wake-up with no externally visible product to assert. Now `@discardableResult` and symmetric; no call-site changes.

iOS suite is 98 tests, 0 failures (was 90). Android suite and both melos gates still clean.

Issue #320, opened 2026-08-07
config: partial setConfig() overwrites the persisted foreground-service config with defaults (showNotificationOnPauseOnly ignored)


GalacticTitan:
## Summary

`setConfig()` is documented and implemented on the native side as a **merge**, but the Dart layer cannot express "I did not set this field". Every `Config` field is non-nullable with a default, so `Config.toMap()` emits a value for every key — and the native merge, which only skips `null`, cannot tell a deliberate `false` from an unmentioned `false`. A `setConfig()` that does not repeat `android.foregroundService` therefore **overwrites the persisted foreground-service configuration with defaults**, and the result is persisted to `SharedPreferences`.

Reported as: *"`showNotificationOnPauseOnly` is true, but when I click start I can clearly see the foreground notification. Also when I click Low-Accuracy GF the foreground notification goes away. On fresh install it works perfectly — it only happens after some time."*

## Why it looks like a regression but isn't

It is latent and reachable only after a `setConfig()` that omits the section. `ready()` re-sends the full config, so a fresh install — and any subsequent Initialize — looks correct. Pressing any demo button that calls a partial `setConfig()` silently rewrites the stored values, and every later `start()` uses them. That is exactly the reported "works on fresh install, breaks after some time" shape.

## Mechanism

`ForegroundServiceConfig` declares non-nullable fields with defaults
(`packages/tracelet/lib/src/models/android_config.dart`):

```dart
const ForegroundServiceConfig({
  this.enabled = true,
  this.channelId = 'tracelet_channel',
  this.notificationTitle = 'Tracelet',
  this.showNotificationOnPauseOnly = false,
  ...
});
```

`toMap()` emits all of them unconditionally, so `const Config()` serialises a
complete foreground-service section made entirely of defaults.

`ConfigManager.setConfig()` (`sdk/android/.../ConfigManager.kt`) merges it in:

```kotlin
val fgService = flat["foregroundService"]
if (fgService is Map<*, *>) {
    for ((k, v) in fgService as Map<String, Any?>) {
        if (v != null) merged["fg_$k"] = v   // `false` is not null → it is written
    }
}
```

The null-skip is the right guard — the comment a few lines below states the
intent explicitly ("a partial `setConfig()` must not overwrite existing non-null
config with defaults") — but Dart never sends `null` for these fields, so the
guard never fires. `persistToPrefs(merged)` then makes the clobbering survive
process death.

## Reproduction

[...]

PR #360, opened 2026-08-12
feat(notification): OS-rendered elapsed timer, targeted notification updates

### LINKED ISSUE #376: Notification: no OS-rendered elapsed timer, and no way to update the notification without a full setConfig
## Problem

The ongoing notification the SDK already shows while tracking cannot do two things:

1. **It cannot display a running timer.** There is no way to tell it "count up from this moment", so an app that wants to show how long the current tracking session has been running has to rewrite the text itself, once a minute, for the whole session.
2. **It cannot be updated quietly.** Every update alerts the device again, so a status timer refreshed once a minute makes the phone sound and vibrate once a minute, for as long as tracking runs.

Both are one builder call away on Android — `setUsesChronometer` and `setOnlyAlertOnce` — and neither is reachable through `ForegroundServiceConfig` today. iOS has the same gap: a Live Activity can render `Text(timerInterval:)` natively, but nothing in the config can ask for it.

For context, this is not a notification an app opts into. A `location`-type foreground service must post a visible, non-dismissible notification for as long as it runs, so every Android tracking app is already occupying that space for the entire session. On iOS the equivalent is the Live Activity, where an app has opted into one. The only question is whether that space says anything useful.

This matters most for apps that turn tracking on and off around a defined activity: a delivery run, a job, a drive, a booked period of work. An app that simply tracks continuously has no such period and will not want a timer — which is why it should be opt-in. "Since tracking started" is only meaningful when tracking is bounded.

## The workaround, and why it is not good enough

Rebuild a full `Config` and call `setConfig()` on a timer. Three costs, and the third is the one that hurts:

1. A full native config write every minute, for the life of the session, to animate a number the OS can render for free.
2. Every update alerts the device again. Changing the text means calling `notify()` with the same notification id — the notification is updated in place, not duplicated, so nothing stacks in the shade. But Android still treats each `notify()` as a fresh alert unless told otherwise, and `setOnlyAlertOnce` is never called here. So the phone sounds and vibrates on every single update, as though a brand-new notification had arrived — once a minute, indefinitely.
3. `Config` carries `http.url` and `http.headers`. Rebuilding the whole object to change a string puts the sync target in scope of a cosmetic update. We lost uploads that way downstream: a config rewrite dropped the auth header mid-session, every POST 401'd, and because the client had already been answered `200` the queued points were discarded. Nothing in the API forced that mistake — but nothing prevented it either.

## Suggested shape

Three additive changes, all default-off:

- `ForegroundServiceConfig.notificationStartedAt` (epoch ms) and `notificationShowTimer` — Android sets `setWhen` + `setUsesChronometer(true)` + `setShowWhen(true)`; iOS renders `Text(timerInterval:countsDown: false)` in the Live Activity. Both tick without any repost.
- `Tracelet.setNotification({title, text, startedAt, showTimer})` — updates only the notification and reposts it, without a `Config` round trip, via a payload that is structurally incapable of carrying `http.*`.
- `notificationOnlyAlertOnce` — exposes `setOnlyAlertOnce`, so an ongoing notification can be updated in place without the device alerting each time.

Happy to open a PR for this.

JoelCouchman:
Fixes #376

## Problem

The ongoing notification the SDK already shows while tracking cannot do two things:

1. **It cannot display a running timer.** There is no way to tell it "count up from this moment", so an app that wants to show how long the current tracking session has been running has to rewrite the text itself, once a minute, for the whole session.
2. **It cannot be updated quietly.** Every update alerts the device again, so a status timer refreshed once a minute makes the phone sound and vibrate once a minute, for as long as tracking runs.

Both are one builder call away on Android — `setUsesChronometer` and `setOnlyAlertOnce` — and neither is reachable through `ForegroundServiceConfig` today. iOS has the same gap: a Live Activity can render `Text(timerInterval:)` natively, but nothing in the config can ask for it.

For context, this is not a notification an app opts into. A `location`-type foreground service must post a visible, non-dismissible notification for as long as it runs, so every Android tracking app is already occupying that space for the entire session. On iOS the equivalent is the Live Activity, where an app has opted into one. The only question is whether that space says anything useful.

This matters most for apps that turn tracking on and off around a defined activity: a delivery run, a job, a drive, a booked period of work. An app that simply tracks continuously has no such period and will not want a timer — which is why it is opt-in. "Since tracking started" is only meaningful when tracking is bounded.

The workaround available today is to rebuild a full `Config` and call `setConfig()` on a timer. Three costs, and the third is the one that hurts:

1. A full native config write every minute, for the life of the session, to animate a number the OS can render for free.
2. Every update alerts the device again. Changing the text means calling `notify()` with the same notification id — the notification is updated in place, not duplicated, so nothing stacks in the shade. But Android still treats each `notify()` as a fresh alert unless told otherwise, and `setOnlyAlertOnce` is never called here. So the phone sounds and vibrates on every single update, as though a brand-new notification had arrived — once a minute, indefinitely.
3. `Config` carries `http.url` and `http.headers`. Rebuilding the whole object to change a string puts the sync target in scope of a cosmetic update. We lost uploads that way downstream: a config rewrite dropped the auth header mid-session, every POST 401'd, and because the client had already been answered `200` the queued points were discarded. Nothing in the API forced that mistake — but nothing prevented it either.

## Approach

Three additive changes, all default-off.

**A timer the OS renders.** `ForegroundServiceConfig` gains `notificationStartedAt` (epoch ms) and `notificationShowTimer`. Android sets `setWhen` + `setUsesChronometer(true)` + `setShowWhen(true)`; iOS renders `Text(timerInterval:countsDown: false)` in the Live Activity. Both tick without any repost.

`notificationStartedAt` is supplied by the app rather than taken from `start()`, deliberately: the session the user cares about often began before tracking did, or survives a tracking restart. An app that wants "since tracking started" simply passes the moment it called `start()`.

The semantics agree across platforms:

- `showTimer` false or unset renders exactly today's notification.
- `showTimer` true with no `startedAt` shows no timer and logs why. Substituting "now" would restart the clock on every repost and misreport the session.
- A `startedAt` in the future is clamped to now **per post**, leaving the persisted value untouched, so a device clock correction self-heals on the next repost rather than being baked in.
- Reposts never restart the clock: the persisted `startedAt` is the only source of truth. This is what makes the timer survive a config change, a visibility flip, or the app process being killed while the service keeps running.
- Count-up only. A terminal Live Activity never ticks.

**`Tracelet.setNotification({title, text, startedAt, showTimer})`.** Updates only the notification and reposts it, without a `Config` round trip. The payload is a distinct Pigeon type that is structurally incapable of carrying `http.*` — the guarantee, rather than a convention.

**`notificationOnlyAlertOnce`.** Exposes `setOnlyAlertOnce`, so the foreground-service notification can be updated in place without the device sounding and vibrating each time. Useful well beyond the timer: any app that reflects live state there — pending uploads, current geofence, accuracy — hits this today.

## A note on the alert-once default

It defaults to `false`, preserving today's behaviour exactly, because everything else here is strictly additive.

Arguably it should default `true`: alerting the device afresh every time the text of an ongoing, non-dismissible notification changes is rarely what anyone wants, and Android's own guidance suggests it for ongoing and progress notifications. But flipping it changes behaviour for every existing app on upgrade, which felt like your call rather than ours. It is a one-word change if you want it.

## Compatibility — one required release step

**The native packages must ship with their `tracelet_platform_interface` constraint raised to the version containing these fields.**

Pigeon serialises `ForegroundServiceConfig` as a positional list. This change appends three entries (indices 12–14), and the generated native decoders read those positions unconditionally:

```kotlin
val notificationStartedAt = pigeonVar_list[12] as Long?
val notificationShowTimer = pigeonVar_list[13] as Boolean?
val notificationOnlyAlertOnce = pigeonVar_list[14] as Boolean?
```

Your release process already handles this — `tracelet_android`, `tracelet_ios` and `tracelet_web` moved to `tracelet_platform_interface: ^3.8.2` with the 3.8.2 release — so this is a "keep doing that" note rather than a new requirement. It matters more than usual here: if these packages ever released while the constraint still admitted an interface version encoding only 12 entries, a mixed resolution (new native package, older interface) would read past the end of the list and throw during `ready()`, on the first call the app makes.

A clean install resolves to matching versions and is fine; the exposure is a pinned interface version or a partially updated lockfile. Preserving the codec type IDs does not help here: the type ID identifies which class is being decoded, not how many fields it has.

We have not touched the constraints, since the correct value is the version this lands in.

Everything else is additive: the new keys are absent unless set, so `ConfigManager` on both platforms never sees them and existing callers are byte-for-byte unaffected. No public Dart signature changed, no default altered, no required parameter added. `TraceletPlatform.setNotification` has an `UnimplementedError` default so third-party platform packages still compile, and the web package no-ops.

Two smaller notes:

- The three new parameters are appended to the end of the native Kotlin `ForegroundServiceConfig` data class, so positional construction, `copy` and `componentN` are unchanged.
- `ContentState` gains the new fields with backward-compatible decoding, so a Live Activity **push** payload in the old shape still decodes after an app upgrade — ActivityKit requires push fields to match the `ContentState` type, so an added non-optional field would otherwise break existing push backends.

## Testing

- `dart analyze --fatal-infos` across all nine packages: clean.
- `packages/tracelet`: 523 passing. The 24 failures are all `flutter_rust_bridge has not been initialized` — the Rust core is not built in this environment — and a clean checkout of `main` fails identically. No assertion failures.
- `packages/tracelet_platform_interface`: 112 passing, 10 failing, identical to a clean checkout.
- `packages/tracelet_android`, `packages/tracelet_ios`: all passing.
- Pigeon regenerated with the pinned generator (26.3.4) and formatted; the generated files are committed.

New coverage: default-off behaviour on both platforms, explicit `false` overriding a stored `true`, the future-timestamp clamp, epoch millis above `Int32` surviving JSON persistence (and a small unsuffixed integer, which is the regression an `as? Long` read would reintroduce), `setNotification` leaving `http.url` intact, and a Live Activity content-state decoding from a pre-upgrade payload.

**Not run: `swift test` and the iOS example build.** Both are macOS-only and this was developed on Windows, so the iOS half is compile-unverified — the Swift changes have had review but not a compiler. Worth your eyes there specifically, and happy to iterate if CI disagrees.


GalacticTitan:
@JoelCouchman please check the tests are failing.

GalacticTitan:
if possible, can you create one github issue and related issue card in the example app like current procedures?.

GalacticTitan:
Thanks for this — the writeup is unusually thorough, and the design calls are the ones I'd have wanted: app-supplied `startedAt` rather than inferring it from `start()`, clamping a future value per-post instead of persisting the clamp, and a separate Pigeon payload that is structurally incapable of carrying `http.*`. Keeping `onlyAlertOnce` defaulting to `false` was also right — let's not fold a behaviour change for existing apps into an additive feature.

Three things to sort before review. One of them is our fault: the workflow below was never written down anywhere in the repo, so there was no way for you to know. I'm fixing that in `CONTRIBUTING.md` now.

## 1. CI is red — `Build Android`

[`:tracelet-sdk:testDebugUnitTest`](https://github.com/Ikolvi/Tracelet/actions/runs/31553069454/job/94057998423) fails, 2 of 378:

```
ApplyChronometerTest > timer disabled leaves builder untouched FAILED
    java.lang.AssertionError at ApplyChronometerTest.kt:29
ApplyChronometerTest > missing startedAt leaves builder untouched FAILED
    java.lang.AssertionError at ApplyChronometerTest.kt:38
```

Both *positive* tests pass, so `applyChronometer` itself looks correct — I think this is the assertions rather than the production code. The two that fail are exactly the two asserting on key presence:

```kotlin
assertFalse(notification.extras.containsKey(Notification.EXTRA_SHOW_CHRONOMETER))
```

`NotificationCompatBuilder` calls `setUsesChronometer(b.mUseChronometer)` unconditionally when it builds the platform notification, so `EXTRA_SHOW_CHRONOMETER` should be present in `extras` either way — `false`, not absent. If that's it, asserting the value instead of the key is the whole fix:

```kotlin
assertFalse(notification.extras.getBoolean(Notification.EXTRA_SHOW_CHRONOMETER))
```

That's a read of the log rather than a verified diagnosis — you'll confirm or kill it faster than I will.

Worth flagging that this suite does **not** run under `melos run test`, which is how it can be green locally and red here. Both Android suites gate every PR:

```bash
# The Kotlin tests load the host Rust core via JNA — build it first.
cd sdk/rust-core && cargo build && cd ../..

cd example/android && ./gradlew :tracelet_android:testDebugUnitTest
cd ../../sdk/android && ./gradlew :tracelet-sdk:testDebugUnitTest :tracelet-sync-sdk:testDebugUnitTest
```

And `cd sdk/ios && swift test` for the Swift side — note that a test file not listed in the `sources:` array in `sdk/ios/Package.swift` is silently never compiled or run, and the build still reports success. Please check your two new iOS test files actually appear in the output.

## 2. There's no issue behind this

Could you open one and add a `Fixes #N` line to the PR body? Your "Problem" section is already the issue text almost verbatim.

This isn't bookkeeping. The issue number is what `git blame` lands on years later, what the `(#N)` comments in the code point at, and what names the verification card below. A PR without one leaves that trail ending in nothing.

## 3. The example app needs a verification card

Every merged change in this repo gets one: a panel in the example app's **Issues** tab that exercises the behaviour on a real device. It matters more than usual here — a chronometer that keeps ticking across a repost, and a Live Activity rendering `Text(timerInterval:)`, are precisely the things a unit test cannot observe. Right now nobody can confirm this feature works without rebuilding your setup by hand.

- Add `example/lib/issues/issue_<N>_card.dart` using `IssueCardShell`
- Register it at the **top** of the list in `example/lib/issues/recent_issues_tab.dart` (newest first), with the import alongside the others
- Run it on a device and make sure it passes

[`example/lib/issues/issue_357_card.dart`](https://github.com/Ikolvi/Tracelet/blob/main/example/lib/issues/issue_357_card.dart) is a good model — it measures a physical behaviour and is explicit about why the measurement is valid.

One rule about what a card may claim: **state only what it actually observes.** If the card can confirm the timer did not restart across a repost but cannot confirm what the lock screen drew, say exactly that in the status output and report the rest as context. A card reporting PASS for something it never measured is worse than no card, because it will be believed.

---

None of this is a comment on the quality of the change — the feature is one we want and the reasoning behind it is sound. Ping me once CI is green and the card is in, and I'll review properly.

JoelCouchman:
Thanks — all three are in, and the workflow note was fair: none of it was written down where I could find it.

## 1. CI

The two failures were the assertions, not the code — your read of the log was right. `NotificationCompatBuilder` calls `setUsesChronometer(mUseChronometer)` unconditionally, so `EXTRA_SHOW_CHRONOMETER` is always present as `false` rather than absent, and a `containsKey` assert could never pass. Confirmed on a device: `hasKey=true value=false`.

They now assert the value, plus that `when` was moved to neither `startedAt` nor the supplied `now`, which is what "leaves the builder untouched" actually means. I did not add an exact-default assertion: `NotificationCompat.Builder` stamps creation time into `when`, so there is no fixed value to compare against, and the comment says that rather than implying more.

Ran the suites you listed. `:tracelet-sync-sdk` is green. `:tracelet_android` has 8 failures on my machine — database, geofence and query tests — but they fail identically on a clean checkout of `main`, and the `:tracelet-sdk` suite fails ~106 the same way, so I have treated the lot as environmental (Windows) rather than anything this branch does. Flagging it only so you know what I saw; I have no reason to think your CI shows the same.

**`swift test` I could not run** — no Mac. The Swift half of this change has had review but never a compiler. Both new iOS test files are listed in `sdk/ios/Package.swift` `sources:`, since you warned that files missing from that array are silently never compiled.

## 2. Issue

#376, opened before this comment. The PR body now carries `Fixes #376`, and the card is `issue_376_card.dart`.

The branch name predates `CONTRIBUTING.md` and so does not follow `feat/<N>-...`. Happy to rename if you would rather, but I did not want to move a branch under an open PR without asking.

## 3. Verification card

`example/lib/issues/issue_376_card.dart`, registered at the top of `recent_issues_tab.dart`, built on `IssueCardShell`, and run on a device.

**One thing to look at first, because it is a change to your SDK and not just the example app.** `applyChronometer`'s applied branch now emits a single debug line:

```
chronometer applied: when=… startedAt=… clamped=…
```

I would rather explain that than have you find it. Your rule — *a card reporting PASS for something it never measured is worse than no card, because it will be believed* — turned out to be much harder to satisfy than I expected:

- `getLogs()` is the only native-observable channel. Pigeon exposes `getState`/`getLogs`/`clearLogs`; there is no `getConfig`, and `getState().config` is backfilled from the Dart `_currentConfig` mirror that `setNotification` itself mutates before calling native.
- My first card read exactly that mirror. It asserted that a Dart field held what the card had just written to it, passed on a device, and proved nothing about your code.
- The chronometer path logs nothing on success — only the degenerate missing-`startedAt` case — so without that line there is no honest PASS row to write.

The call site passes `configManager.getFgNotificationStartedAt()`, so the logged value is the one the native config held while configuring the builder, which is what makes the line evidence rather than an echo of the call. The same line across a text-only repost is what lets the card show the rebuild was configured from the same instant — your own example of a claim worth making. **If you would rather not carry a log line, say so and I will drop it; the card then ships with no native rows and an explicit statement that nothing is natively verified.**

What the card deliberately does not claim is the part I would most like checked:

- The log is written before `builder.build()` and before `notify()`, so no row says anything was built, posted or rendered. It proves the builder's setters received those values.
- Every row is tagged by evidence kind — native log, native State, or Dart mirror — because they are not the same thing and the summary should not pretend otherwise.
- Whether values were persisted is **not** measured. `getFgNotificationStartedAt()` reads an in-memory `configCache`, not SharedPreferences.
- What the shade drew, whether the device alerted, and the iOS Live Activity are reported for the operator to look at, never asserted.

Two limitations worth naming rather than leaving for you to hit:

- The card leaves `notificationStartedAt` set and tracking running. There is no API to clear the field — null fields are merge-skipped — and the notification is deliberately left up so the operator can look at it. A later card can therefore inherit a stale start instant and a `start()` no-op. The card's closing text says so, and the failure path says it too. Happy to make it tear everything down instead if you would prefer that trade.
- It ran on an **emulator**, not physical hardware.

It went through seven passes and two independent reviewers before I was willing to post it; the first four versions all passed on a device while claiming things they had not measured. If it still overclaims anywhere, I would genuinely like to know — that is the failure mode I have been least able to see in my own work.

GalacticTitan:
Merged `main` in and resolved the two conflicts, but I also reduced the scope before landing this — flagging it here rather than leaving it to be discovered in the diff.

**Kept, whole:** the timer. `setUsesChronometer`, `setOnlyAlertOnce` and `setShowWhen` appear nowhere on `main`, so the gap you describe is real, and `applyChronometer` / `applyOnlyAlertOnce` are untouched, along with `notificationStartedAt`, `notificationShowTimer`, `notificationOnlyAlertOnce`, the Live Activity `startedAt`/`showTimer`, and their tests. The alert-once default stays `false` as you left it.

**Removed:** `Tracelet.setNotification()` and `TlNotificationUpdate`.

The reasoning is the third cost in your write-up — that `Config` carries `http.url`/`headers`, so a cosmetic update puts the sync target in scope. That stopped being true with #320/#321. `setConfig` is a partial update: it deliberately sends the fields you supplied rather than the merged object, and `config_partial_setconfig_test.dart` already asserts `http.url` survives a `foregroundService`-only write. Its own docstring uses exactly that example. So the incident you hit downstream was a full-`Config` rebuild in app code rather than something the API forced — and today it doesn't force it.

That leaves `setNotification()` buying one call in place of `setConfig` + `updateNotification()`, for a Pigeon type, a host-API method across Android/iOS/web plus the platform-interface default, and a fourth way to update the same notification. Removing it also retires the mixed-version hazard you flagged: with `TlNotificationUpdate` gone, `TlLocationTuning` is the last-declared class again and no codec type IDs move.

Your #376 card keeps every row — it reads `applyChronometer`'s log lines, which didn't change. It now drives them through a local `_setNotification` helper that writes only `foregroundService` fields through `setConfig` and reposts. The http row is arguably stronger for it: it now pins the partial-merge guarantee itself, across all six writes, rather than a bespoke payload's shape.

Two notes back to you:

- The iOS half is no longer compile-unverified. It builds for iOS and the suite runs green on an iPhone 17 simulator — 115 tests, 0 failures, including your `ConfigManagerLiveActivityTimerTests` (3) and `TraceletActivityAttributesDecodingTests`. Both new files were correctly added to `Package.swift`'s `sources:`, which is the trap that would otherwise have let them silently never run.
- Your release note stands and still needs doing: the native packages pin `tracelet_platform_interface: ^3.8.3`, which resolves to an interface encoding 12 `foregroundService` entries, and the regenerated decoders read indices 12–14 unconditionally. That constraint has to move to whatever version this lands in.

Also verified green: `melos analyze` 12/12, `melos format`, `tracelet` 543 tests, `tracelet_platform_interface` 311, and the Android SDK and plugin Kotlin suites. Thanks for the chronometer work and for the unusually careful card — the honesty about what the log line does and doesn't evidence made this much faster to review.

Issue #361, opened 2026-08-12
maxDaysToPersist / maxRecordsToPersist accepted but never enforced (Android, 3.7.6 & 3.8.0)


jsuarezarrojo:
## Summary

`maxDaysToPersist` and `maxRecordsToPersist` (`PersistenceConfig`) are accepted by `ready()`/`setConfig()` and echoed back correctly in the returned `State.config`, but neither limit is ever enforced against the `location_events` table. The local queue grows without bound regardless of the configured caps.

`maxRecordsToPersist` confirmed on **Android**, on both `tracelet: 3.7.6` / `tracelet_sync: 3.7.6` and `tracelet: 3.8.0` / `tracelet_sync: 3.8.0` (re-tested on 3.8.0 specifically because the 3.8.0 changelog doesn't mention retention/pruning, to rule out an already-fixed regression rather than assume). `maxDaysToPersist` confirmed on 3.7.6 only — see the note under Actual. Not tested on iOS or web, but the Swift side has an analogous unused field (see below) so it may not be Android-only.

## Steps to reproduce

1. `Tracelet.ready(Config(persistence: PersistenceConfig(maxRecordsToPersist: 3, maxDaysToPersist: -1), http: HttpConfig(autoSync: false, ...)))`.
2. Confirm the cap was accepted: `state.config.persistence.maxRecordsToPersist == 3`.
3. Insert more than 100 explicit locations via `Tracelet.insertLocation()` (unique UUIDs, `autoSync: false` so nothing drains the queue via delivery).
4. Take snapshots with `Tracelet.getPendingLocationCount()` at intervals crossing 100 inserts (e.g. at 99, 100, 101, 105).

## Expected

The pending queue never exceeds 3 records (the configured `maxRecordsToPersist`), with the oldest records purged as new ones arrive.

## Actual

The queue grows unbounded past the cap. Observed checkpoints (inserts → pending count):

| Inserts | Pending count |
|---:|---:|
| 99 | 100 |
| 100 | 101 |
| 101 | 102 |
| 105 | 106 |

Same shape of `maxRecordsToPersist` failure on both 3.7.6 and 3.8.0.

[...]

Issue #356, opened 2026-08-11
[Feature]: Automatic HTTP Sync for Trips and Telematics Events


sfaizanh:
### Is your feature request related to a problem?

No

### Summary

I'm currently using Tracelet in a Flutter application for continuous background location tracking and are now expanding the implementation to support complete trip tracking and driving telematics events such as:

- Harsh braking
- Harsh acceleration
- Harsh cornering
- Speeding

Our location records are already automatically synced using `HttpConfig` with `autoSync: true`.

We would like to understand whether **completed trips and telematics events can also be automatically synced to our backend**, similar to location records.

### Current Setup

We currently configure Tracelet with an HTTP endpoint similar to:

```dart
http: tl.HttpConfig(
  url: 'https://api.example.com/users/coordinates-telemetry',
  method: tl.HttpMethod.post,
  autoSync: true,
  batchSync: true,
  maxBatchSize: 50,
),
```

This works well for background locations, including offline persistence and retry.

We are now planning to enable:

```dart
telematics: tl.TelematicsConfig(
  enableDrivingEvents: true,
)
```

and use:

```dart
Tracelet.onDrivingEvent(...)
Tracelet.onTrip(...)
```

### Desired Behavior

Ideally, Tracelet would provide automatic persistent HTTP synchronization for all three types of telemetry:

```text
Locations
    ↓
Automatic persistence + retry + HTTP sync

Telematics Events
    ↓
Automatic persistence + retry + HTTP sync

Completed Trips
    ↓
Automatic persistence + retry + HTTP sync
```

For example, a completed trip could automatically produce a payload containing:

```json
{
  "id": "uuid",
  "type": "trip",
  "distance": 18420,
  "duration": 1320,
  "averageSpeed": 13.95,
  "startLocation": {},
  "stopLocation": {},
  "waypoints": []
}
```

and a driving event could produce:

```json
{
  "type": "harsh_braking",
  "severity": 0.82,
  "value": 0.47,
  "speed": 18.4,
  "latitude": 24.8607,
  "longitude": 67.0011,
  "timestamp": "..."
}
```

### Questions

1. Are `DrivingEvent` records currently automatically synced through Tracelet's HTTP sync engine, or are they only persisted locally and exposed through `onDrivingEvent()` / `getTelematicsEvents()`?

2. Is there currently a configuration option such as `syncTelematics` that enables automatic HTTP upload of telematics events? If so, could you provide an example for Flutter?

[...]

Issue #313, opened 2026-08-07
telematics: getTelematicsEvents() returns the oldest unsynced rows, not the most recent history


GalacticTitan:
## Summary

`Tracelet.getTelematicsEvents(limit)` is documented as returning the most recent stored driving/crash events, but the underlying query returns the **oldest** events and **only the unsynced** ones.

## Detail

`sdk/rust-core/core/src/database/mod.rs`:

```rust
let mut stmt = conn.prepare("SELECT id, event_type, severity, latitude, longitude, timestamp, synced FROM tracelet_telematics WHERE synced = 0 ORDER BY id ASC LIMIT ?1")
```

Two consequences for the public history API:

1. **Synced events disappear.** `markTelematicsSynced` is called on both platforms after a successful custom-path sync, and the default payload path calls `clearTelematicsEvents()`. Once an app enables `syncTelematics`, its local history empties out.
2. **`ORDER BY id ASC` + `LIMIT` returns the oldest N**, not the newest.

Both consumers promise the opposite:

- Docs: *"Fetch the 50 most recent telematics events from the database"*
- `packages/tracelet_doctor/lib/src/bug_report.dart`: `## Telematics events (most recent)`

## Why it is written that way

The same function serves the sync batcher (`getTelematicsForCustomBuilder`, and the default `__telematics` injection), which genuinely wants unsynced-oldest-first so a batch can be marked synced up to a max id. That is correct for sync and wrong for history.

## Expected

Split the two: keep the existing query for the sync path, and give the public history API its own query — newest-first, regardless of `synced`.

GalacticTitan:
Fixed in #315, merged to `main` as 9fdad1ea and released in 3.8.0.

(Not auto-closed on merge: the PR body listed these as `Fixes #309, #310, …`, and GitHub only applies the keyword to the first issue in a comma-separated list.)

PR #369, opened 2026-08-12
fix(telematics): stop losing driving events on sync, persist their magnitudes, honor telematicsUrl

### LINKED ISSUE #366: Telematics sync: a failed POST still deletes the events, and a successful one clears the whole table
### What happens

With `http.syncTelematics: true`, a sync attempt that **fails** still deletes every telematics event from the database. An offline device does not queue its driving events for retry — it destroys them.

Separately, even on a successful sync the code deletes the *entire* table rather than the range it actually uploaded, so any event recorded between the batch read and the POST completing is destroyed without ever being sent.

### Mechanism

`syncBatch` attaches unsynced telematics to the location batch and sets `telematicsCleared = true` (Android `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt:2321-2341`):

```kotlin
var telematicsCleared = false
if (syncTelematics) {
    val telematics = db.getTelematicsEvents(250)
    if (telematics.isNotEmpty()) {
        ...
        newExtras["__telematics"] = jsonArray.toString()
        configHttp = configHttp.copy(extras = newExtras)
        telematicsCleared = true
    }
}
```

Then the post-sync branch (`TraceletSdk.kt:2350-2363`):

```kotlin
val count = provider.syncBatchBlocking(configHttp, records)
if (count > 0L || hasTelematics) {          // ← entered even when the POST failed
    if (count > 0L) {
        // ... location cleanup, correctly gated on count
    }
    if (telematicsCleared) {
        db.clearTelematicsEvents()          // ← unconditional DELETE of all rows
    }
}
```

`count` is `0` on HTTP failure — `NativeSyncProvider.syncBatchBlocking` returns `0L` and emits `"success" to false, "responseText" to "Sync failed"` (`sdk/android/tracelet-sync-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/sync/NativeSyncProvider.kt:383-390`). But `hasTelematics` is `true`, so the guard passes anyway, and `clearTelematicsEvents()` runs — `DELETE FROM tracelet_telematics` with no predicate (`sdk/rust-core/core/src/database/mod.rs:1020-1025`).

iOS has the identical structure at `sdk/ios/Sources/TraceletSDK/TraceletSdk.swift:1862-1876`.

### Why the wholesale delete is wrong even on success

`mark_telematics_synced(max_id)` exists for exactly this purpose (`sdk/rust-core/core/src/database/mod.rs:1012`), and the custom-sync-body path already uses it correctly via `markExposedTelematicsSynced()` (`TraceletSdk.kt:2719`). Only the default payload path takes the destructive shortcut.

The sync query is `WHERE synced = 0 ORDER BY id ASC LIMIT 250`, so the batch is a bounded prefix — the correct cleanup is `markTelematicsSynced(batch.last().id)`, not a table-wide delete.

This also still defeats #313. That issue split the history view (`get_telematics_history`, newest-first, ignores `synced`) from the sync view specifically so that enabling `syncTelematics` would stop emptying the app's own local history — but `clearTelematicsEvents()` deletes the rows outright, so `Tracelet.getTelematicsEvents()` still goes empty after every sync.

### Reproduction

1. Configure with `telematics: TelematicsConfig(enableDrivingEvents: true)` and `http: HttpConfig(url: ..., autoSync: true, syncTelematics: true)`.
2. Drive/simulate until several harsh-braking events are recorded; confirm `Tracelet.getTelematicsEvents(10)` returns them.
3. Put the device in airplane mode (or point `url` at a host that refuses the connection).
4. Trigger a sync.

**Expected:** the POST fails, events stay in the table, and the next successful sync uploads them.
**Actual:** the POST fails and the events are gone. `getTelematicsEvents(10)` returns empty, and the backend never receives them.

### Fix

- Gate telematics cleanup on the sync having actually succeeded, not on telematics having been attached.
- Replace `clearTelematicsEvents()` with `markTelematicsSynced(maxId)` for the id range that was uploaded.
- Apply on both Android and iOS.

Reported while investigating #356, which asks for reliable telematics/trip sync — this is the opposite of that guarantee.

### LINKED ISSUE #367: Persisted telematics events drop speed and value: stored history and synced payloads lose the magnitudes
### What happens

A `DrivingEvent` carries `speed` and `value`, and both reach the app through `Tracelet.onDrivingEvent()`. Neither is persisted. Any consumer that reads stored history (`Tracelet.getTelematicsEvents()`, the Doctor bug report) or receives synced telematics gets an event with no speed and no magnitude — so a harsh-braking event arrives with no indication of how hard the braking was or how fast the vehicle was going.

### Mechanism

The event model has seven fields (`sdk/rust-core/core/src/algorithms/telematics.rs:61-74`):

```rust
pub struct DrivingEvent {
    /// `harsh_braking` | `harsh_acceleration` | `harsh_cornering` | `speeding`.
    pub kind: String,
    /// Normalized 0–1 severity (how far past the threshold).
    pub severity: f64,
    /// Speed at the event (m/s).
    pub speed: f64,
    /// The measured magnitude that triggered it: g for harsh events, km/h over
    /// the limit for speeding.
    pub value: f64,
    pub latitude: f64,
    pub longitude: f64,
    pub timestamp_ms: i64,
}
```

The callback delivers all of them (`sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt:3369-3379`), but the very next statement persists only four (`TraceletSdk.kt:3383`):

```kotlin
rustDatabase?.insertTelematicsEvent(e.kind, e.severity, e.latitude, e.longitude)
```

`speed` and `value` are dropped on the floor. The table has no columns for them (`sdk/rust-core/core/src/database/mod.rs:193-203`):

```sql
CREATE TABLE IF NOT EXISTS tracelet_telematics (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    event_type TEXT NOT NULL,
    severity REAL NOT NULL,
    latitude REAL NOT NULL,
    longitude REAL NOT NULL,
    timestamp TEXT DEFAULT (datetime('now')),
    synced INTEGER DEFAULT 0
)
```

iOS drops them at the same point (`sdk/ios/Sources/TraceletSDK/TraceletSdk.swift:3288`, `:3313`).

### Why it matters

`severity` alone is not a substitute. It is documented as normalized 0–1 "how far past the threshold" — it tells you an event was over the line but not the physical quantity. `value` is the actual measurement (g, or km/h over the limit) and `speed` is the context that makes it interpretable. A telematics/insurance backend scoring driver behaviour needs the magnitudes, not just a normalized flag.

This is also the practical blocker for #356: the requester's example driving-event payload asks for exactly `severity`, `value`, `speed` — and `value` and `speed` are the two we cannot supply from storage.

Note this is only visible once you leave the live callback. In-process listeners on `onDrivingEvent()` see the full event, which is why the gap has gone unnoticed.

### Reproduction

1. Enable `telematics: TelematicsConfig(enableDrivingEvents: true)`.
2. Register `Tracelet.onDrivingEvent((e) => print(e))` and generate a harsh-braking event.
3. Observe the callback receives `speed` and `value` populated.
4. Call `Tracelet.getTelematicsEvents(10)`.

**Expected:** the stored record carries the same `speed` and `value`.
**Actual:** the stored record has only `event_type`, `severity`, `latitude`, `longitude`, `timestamp`, `synced`. The magnitudes are unrecoverable.

### Fix

- Add `speed REAL` and `value REAL` columns to `tracelet_telematics`, with `ALTER TABLE ... ADD COLUMN` migrations alongside the existing ones so upgrades are seamless (old rows read back as `0`/null).
- Extend `insert_telematics_event` and `DbTelematicsRecord` to carry both.
- Pass them through at both call sites (Android + iOS) and include them in the synced payload and the custom-builder map.

This changes `#[uniffi::export]` surface, so the three committed binding files need regenerating.

Found while investigating #356.

### LINKED ISSUE #368: HttpConfig.telematicsUrl is accepted and plumbed to the Rust core, but never read — telematics always go to HttpConfig.url
### What happens

`HttpConfig.telematicsUrl` is a public, documented config field that is fully plumbed from Dart through Pigeon, through both native config managers, and into the Rust config struct — and then never read by anything. Setting it has no effect whatsoever: telematics events still go to `HttpConfig.url` as part of the location batch, and the app gets no error, no warning, and no log line indicating the value was ignored.

Same failure shape as #361 (`maxDaysToPersist` / `maxRecordsToPersist` accepted but never enforced): the config surface promises routing that the engine does not implement.

### The field is real all the way down

Public Dart API, documented as if functional (`packages/tracelet/lib/src/models/config.dart:1556-1557`):

```dart
/// The URL to sync telematics events to.
final String? telematicsUrl;
```

It survives the entire config pipeline:

- Pigeon schema — `packages/tracelet_platform_interface/pigeons/tracelet_api.dart:311`
- Android config manager — `ConfigManager.kt:613`, `TraceletConfig.kt:373`
- iOS config manager — `ConfigManager.swift:347`, `TraceletConfig.swift:483`
- Handed to the Rust core — `TraceletSdk.kt:4339` (`telematicsUrl = configManager.getTelematicsUrl()`), `TraceletSdk.swift:4176`
- Lands in the Rust config struct — `sdk/rust-core/core/src/config/mod.rs:337-339`:

```rust
/// URL to sync telematics data to.
pub telematics_url: Option<String>,
```

And that is the end of it. `telematics_url` appears exactly once in the entire Rust codebase — that field declaration. There is no read, no routing decision, no second HTTP call. The sync provider never receives it: `SyncHttpConfig` has no telematics field at all, so it is not even reachable from the code that performs the POST.

The actual behaviour is unconditional — when `syncTelematics` is on, events are serialized into `extras.__telematics` and posted to `HttpConfig.url` along with the locations (`TraceletSdk.kt:2336-2338`, `TraceletSdk.swift:1843-1845`). Nothing consults `telematicsUrl` on the way.

### Reproduction

```dart
http: tl.HttpConfig(
  url: 'https://api.example.com/locations',
  autoSync: true,
  syncTelematics: true,
  telematicsUrl: 'https://api.example.com/telematics', // ignored
),
```

Generate a driving event and sync.

**Expected:** telematics events POST to `https://api.example.com/telematics`.
**Actual:** nothing is ever sent to that host. The events arrive at `https://api.example.com/locations` inside `extras.__telematics`. No warning is logged.

A backend team routing telematics to a separate service will see an endpoint that receives zero traffic with no indication why.

### Fix

Implement the routing the field already promises: when `telematicsUrl` is set and non-empty, upload telematics events to that endpoint as their own request rather than folding them into the location payload; when it is unset, keep the current attached-to-locations behaviour so existing integrations are unaffected.

The upload must use the same success/failure accounting as #366 — a failed telematics POST must leave the rows unsynced for retry, and must not take the location batch down with it.

Found while investigating #356, which asks specifically whether separate per-type endpoints are supported.

GalacticTitan:
Three defects in the telematics path, all surfaced while answering the sync questions in #356.

## #366 — a failed sync deleted the events

The post-sync branch keyed off *"telematics were attached"* rather than *"the POST succeeded"*:

```kotlin
val count = provider.syncBatchBlocking(configHttp, records)
if (count > 0L || hasTelematics) {      // ← true even when the POST failed
    if (telematicsCleared) {
        db.clearTelematicsEvents()      // ← DELETE FROM tracelet_telematics
    }
}
```

`syncBatchBlocking` returns `0` on HTTP failure, so an offline device ran straight into an unpredicated `DELETE` and destroyed the driving events it was supposed to be queueing — the exact scenario the reliability request in #356 asks to be protected from.

Two things were wrong; both are fixed:

- Events are settled only when the request that carried them actually succeeded.
- Settling now marks the uploaded id range synced instead of clearing the table, so an event recorded between the batch read and the POST is no longer collateral.

`markTelematicsSynced` already existed and the custom-sync-body path already used it correctly — only the default payload path took the destructive shortcut.

Marking rather than deleting is also what finally makes #313 hold: uploading an event no longer removes it from the app's own history. That leaves the table growing for the lifetime of the install, so the synced tail is trimmed to the newest 1000 rows. **Unsynced rows are exempt from the trim** — they are still owed to the server, and losing them is the bug this accompanies.

## #367 — `speed` and `value` were dropped at the insert

`DrivingEvent` carries both and `onDrivingEvent` delivered both, but the table had no columns for them:

```kotlin
rustDatabase?.insertTelematicsEvent(e.kind, e.severity, e.latitude, e.longitude)
```

So a live listener saw a complete event while anything reading it back — or receiving it through sync — got a normalized 0–1 `severity` flag with no physical quantity behind it. `severity` is not a substitute: it says how far past the threshold the event landed, not how hard the braking was or how fast the vehicle was going.

Added via `ALTER TABLE` alongside the existing migrations, so upgrades are seamless (pre-migration rows read back as `0.0`). Impacts map to their natural analogues: entry speed and peak g. The fields are carried through the Rust core, the uniffi bindings, Pigeon, and the Dart `TelematicsRecord`.

## #368 — `telematicsUrl` was accepted but never read

The field was plumbed from the public Dart API all the way to the Rust config struct, where `telematics_url` appeared exactly once in the entire codebase — its own declaration. It never reached the sync provider at all (`SyncHttpConfig` has no telematics field), so a separate telematics endpoint silently received zero traffic with no warning. Same shape as #361, where the persistence caps were accepted and never enforced.

It now does what it always claimed: when set, telematics are POSTed to that endpoint on their own request as `{"telematics": [...]}`, routed through the sync provider so headers, timeouts, retry/backoff, 401 token refresh and SSL pinning behave exactly as they do for locations. Failure is accounted for separately from the location batch — a failed telematics POST leaves those rows unsynced and does not take the locations down with it.

## Backward compatibility

This was a design constraint throughout, since existing deployments are already receiving telematics:

- **Wire format is additive.** `speed` and `value` join the objects in `extras.__telematics`; no key is renamed or removed.
- **Default routing is unchanged.** Leave `telematicsUrl` unset — the default — and telematics keep riding the location request exactly as before. The new path only engages when an app explicitly asked for it.
- **Third-party sync providers keep compiling.** `postTelematicsBlocking` has a default implementation on both the Kotlin and Swift `SyncProvider` contracts, and it returns "not delivered" rather than settling events it never sent.
- **`simulateTelematicsEvent` keeps its 4-argument signature**, so the Pigeon and React Native surfaces are untouched.
- The new `speed`/`value` fields are nullable on the Dart side, so a Dart layer newer than the native plugin decodes a missing field rather than failing.

## Verification

- `cargo test -p tracelet_core --lib` — 143 passed, including five new tests covering the magnitude round-trip, legacy rows reading back as `0.0`, and the synced-tail trim never taking unsynced rows.
- `./gradlew :tracelet-sdk:compileDebugKotlin :tracelet-sync-sdk:compileDebugKotlin` — clean.
- `flutter test` in `tracelet` (534) and `tracelet_platform_interface` (311) — all passing.
- `melos format` and `melos analyze` — both clean.
- uniffi bindings regenerated from the host dylib; the diff contains only the intended signature changes.

Example issue cards added for all three (`example/lib/issues/issue_366_card.dart`, `_367_`, `_368_`). #366's card drives the real failure path — it points the sync at a refused port and asserts the events survive, twice. #367's and #368's cards state plainly what they can and cannot prove locally: #367 proves the wiring round-trips but cannot synthesize a genuine harsh-braking event, and #368 cannot observe *which* host received a body without a live endpoint.

Fixes #366
Fixes #367
Fixes #368

Groundwork for #356, which stays open as the tracker for the feature itself: trip persistence, trip sync, and route context on telematics and trip records.


GalacticTitan:
Adds **#370**, found while verifying #368 on a device: the card enabled `syncTelematics`, stored a driving event, called `sync()` — and reported **zero HTTP requests**.

`sync()` read the flag as `getConfig()["http"]["syncTelematics"]`, but `ConfigManager.setConfig` flattens Dart's nested sections into the top level on the way in, so `["http"]` was always null and the flag always `false`. **`syncTelematics` has never taken effect on this path on either platform.**

It is silent in both directions — a false flag just skips the block, and `State.config` still reports `true` because that reads a different structure — which is why it survived this long. The custom-sync-body path uses the correct accessor, so apps with `setSyncBodyBuilder` registered did get their telematics, making it look integration-specific rather than broken.

It also partially masked #366: telematics were rarely attached on the default path, so the destructive clear fired less often than it otherwise would have.

Fixed by reading through `getSyncTelematics()` on both platforms, with regression tests in `ConfigManagerSectionFlatteningTest` — already the home for this class of bug (#74). They pin the flat accessor and assert `getConfig()["http"]` is null, so the nested form cannot come back. 46 tests, 0 failures.

Also fixed here: the `records.isEmpty` short-circuit I added for #368 was skipping the provider call on the *default* path too, so telematics attached to an empty location payload would never have been delivered. It now only short-circuits when they already went to their own endpoint.

Issue #383, opened 2026-08-13
[Bug]: PersistMode is not honoured for geofence transitions — location and none both persist and sync geofence rows anyway


JoelCouchman:
## Package Name

tracelet (Core Plugin)

## Tracelet Version

3.8.4

## Platform

Android, iOS

## Device & OS Version

Reproduced by code trace on both platforms; not device-specific. Confirmed by two independent traces (file:line references below) rather than by a single device repro.

## Description / Scenario

`PersistenceConfig.persistMode` controls which record types get written to the
local database. The four values, and what the SDK documents for each:

```dart
enum PersistMode {
  /// Persist all location and geofence records (default).
  all,

  /// Persist only location records.
  location,

  /// Persist only geofence records.
  geofence,

  /// Do not persist any records. Events are still fired.
  none,
}
```

(`packages/tracelet_platform_interface/lib/src/types/enums.dart:241-253`)

This is honoured for ordinary GPS locations — `persistLocationIfAllowed` skips
the DB insert for modes `geofence` (2) and `none` (3):

- Android: `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/location/LocationEngine.kt:1736-1739`
- iOS: `sdk/ios/Sources/TraceletSDK/location/LocationEngine.swift:1741-1744`

It is **not** honoured for geofence ENTER/EXIT transitions, on either
platform. A geofence transition is wired straight from the geofence manager
to the SDK's unconditional insert method, with no persist-mode check in
between:

**Android**
- `GeofenceManager.handleGeofenceEvent` (OS-delivered transitions) and
  `evaluateHighAccuracyProximity` (software-evaluated transitions) both call
  `onGeofenceEvent?.invoke(eventData)` with no mode check —
  `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/geofence/GeofenceManager.kt:677`
  and `:894`.
- `onGeofenceEvent` is wired directly to the SDK's insert method at setup —
  `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt:672-674`:
  ```kotlin
  onGeofenceEvent = { eventMap ->
      insertLocation(eventMap)
  }
  ```
- `insertLocation` (`TraceletSdk.kt:2126`) recognises the `"geofence"` event
  type, writes the row through `db.insertLocation(...)` (`:2242`), and
[...]

Issue #357, opened 2026-08-11
Geofence: a fence added after start() never gets the fix cadence it needs — EXIT fires late or not at all


GalacticTitan:
A fence the in-app evaluator owns (a polygon, or a circle under 100 m) is decided from the raw fix stream, so #356 drops the OS distance filter to give the evaluator a time-based cadence:

```swift
// TraceletSdk.swift — wireGeofenceLocationCallbacks()
locationEngine.geofenceHighAccuracyMode = geofenceManager.hasEvaluatorOwnedGeofences()
```

That question is asked **once**, at `start()` / `startGeofences()` / relaunch. The fence set is mutable afterwards and nothing re-asks it: `addGeofence()`, `addGeofences()`, `removeGeofence()` and `removeGeofences()` all go straight to the manager and never touch the engine.

So the ordinary sequence — start tracking, then add a fence — leaves the flag `false` for the rest of the session:

```dart
await Tracelet.start();                                   // no fences yet -> flag = false
await Tracelet.addGeofence(Geofence(radius: 10, ...));    // never re-asked
```

CoreLocation keeps `distanceFilter = 10` (the configured value) and the evaluator is fed one fix per 10 m travelled. Android is the same shape: `minUpdateDistanceMeters` stays at the configured filter (`LocationEngine.buildLocationRequest()`).

### Field evidence (iOS 26.5.2, 10 m fence, `distanceFilter: 10`)

**Run 1 — no EXIT at all.** Deliveries during the walk were 10.49 m, 10.32 m, 10.90 m apart — the distance filter, not a coincidence. EXIT needs `(dist - acc) > radius + buffer` on two consecutive fixes; the walk produced exactly one qualifying fix (16.8 m, acc 2.1), the next arrived 32 s later at 12.7 m and reset the pending count.

```
16:48:03 [geofence] ENTER issue-356-tiny dist=0.1 radius=10.0 buffer=3.0 thr=13.0
(no EXIT — 3 fixes in 95 s)
```

[...]

Issue #299, opened 2026-08-05
Distance over-reported on foot: odometer ignores Kalman smoothing, and the transport classifier never tunes the filters


GalacticTitan:
Reported on Discord (05 Aug 2026): walking a neighbourhood loop on a Pixel 9, the recorded track shows lateral "spikes" and distance is over-reported by ~3x. After enabling both the location filter and the Kalman filter the user was still ~50% over, and landed on hand-tuned values (`distanceFilter: 5`, `trackingAccuracyThreshold: 10`, `maxImpliedSpeed: 5`) by trial and error. Strava on the same device, same walk, produced a smooth track.

Two independent defects, plus the feature that makes the hand-tuning unnecessary.

## Part 1 — `useKalmanFilter` has no effect on distance (bug)

`LocationProcessor::process` computes `odometer_delta` with `haversine()` over the **raw** fix (`algorithms/location_processor.rs:334-338`, `:404-410`). The Kalman filter runs **after** the processor returns, in the platform layer (`LocationEngine.kt:989-1001`, `LocationEngine.swift` equivalent), and only its output reaches `coords`.

Consequence: turning on `useKalmanFilter` smooths the rendered polyline but leaves `state.odometer` accumulating raw inter-fix noise. A user who enables it to fix their distance sees no change and has no way to tell why. That is precisely the reported symptom.

Fix: smooth first, then filter and accumulate. Feeding the smoothed coordinates into the processor also makes the distance filter, accuracy filter and implied-speed filter operate on the de-noised track, which is the intent everywhere those thresholds are documented.

## Part 2 — the adaptive engine is fed the wrong activity signal (bug)

`LocationEngine.kt:942-948` builds the `AdaptiveContext` from `currentActivityType` — the raw platform Activity Recognition string — rather than `effectiveActivityType()` (`LocationEngine.kt:129-136`). So even with `enableFusedClassifier: true` **and** `fusedClassifierAuthoritative: true`, the fused classifier's mode never reaches `AdaptiveSamplingEngine`. The setting silently does nothing for sampling, despite `fusedClassifierAuthoritative` being documented as "the fused mode overrides the platform activity for sampling" (`driving_config.dart:186-187`).

## Part 3 — auto-tune the filters from the detected mode (feature)

[...]

PR #386, opened 2026-08-14
fix(motion): acquire the first location on a stationary start (#385)

### LINKED ISSUE #385: [Bug]: a session that starts stationary never acquires a first location — nothing is delivered until the device physically moves
## Summary

A continuous session that starts in the **stationary** pace — which is the default, since
`motion.isMoving` defaults to `false` — acquires **no location at all**. Not a delayed one, not a
coarse one: nothing is ever delivered until the accelerometer or activity recognition declares
motion. The app that called `start()` sits with an empty map, and `onLocation` stays silent until
the user physically walks away.

This is the single most-reported "the SDK isn't working" symptom, and the workaround handed out so
far has been to configure `motion: MotionConfig(isMoving: true)`. That is the wrong advice: it
starts a full-rate GPS stream and holds it until `stopTimeout` expires (battery), fires a
`motionchange` the app never asked for, and overrides a documented config value to work around a
missing behaviour.

## Mechanism

`start()` commits the pace and then branches on it. Only the moving branch touches the location
provider.

**Android** — `sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt`:

```kotlin
if (stateManager.isMoving) {
    locationEngine.start()
} else {
    changePace(false)
    // ... only an in-app-evaluated geofence (#357) starts the stream here
    if (locationEngine.geofenceHighAccuracyMode) {
        locationEngine.start()
    }
}
```

**iOS** — `sdk/ios/Sources/TraceletSDK/location/../TraceletSdk.swift`:

```swift
if stateManager.isMoving {
    locationEngine.start()
    startBackgroundActivitySessionIfNeeded()
} else {
    _ = changePace(false)
}
```

`changePace(false)` does not fill the gap, and in SMART mode it cannot even fall back to the
stationary periodic schedule. The sequence on a fresh SMART start is:

1. `syncCurrentMode()` reads the just-committed pace and sets the Rust coordinator's
   `current_mode` to `StationaryPeriodic` (`SmartMotionCoordinator.syncCurrentMode`, keyed on
   `stateManager.isMoving` since #344).
2. `adoptSpeedMotionPace()` pushes `is_speed_moving = false` into the coordinator.
3. `evaluate_state` (`sdk/rust-core/core/src/state/smart_motion_coordinator.rs`) computes
   `should_be_moving = false`, sees `current_mode != Continuous`, and returns
   `CoordinatorAction::None`:

```rust
if should_be_moving && state.current_mode != TrackingMode::Continuous {
    ...
} else if !should_be_moving && state.current_mode == TrackingMode::Continuous {
    ...            // schedules the stationary periodic worker
} else {
    CoordinatorAction::None
}
```

So `handleAction` runs nothing: `PeriodicLocationWorker` is never scheduled, `locationEngine.start()`
is never called, and no one-shot is fired. The session is genuinely dark. `startHeartbeat()` runs but
the heartbeat only re-emits `locationEngine.getLastLocation()`, which is `null`, so it contributes
nothing either.

The one-shot machinery already exists on both platforms — it just isn't reachable from `start()`:

* Android `LocationEngine.requestImmediateFix()` (private), fired from `changePace(true)` on a
  stationary → moving *transition*.
* iOS `locationManager.requestLocation()` in `LocationEngine.changePace(_:)`, same transition.

A session that *begins* stationary never performs that transition, so it never gets the fix.

## Reproduction

```dart
await Tracelet.ready(const Config(
  // no motion.isMoving override — the default, false, is the whole point
  logger: LoggerConfig(logLevel: LogLevel.debug),
));
Tracelet.onLocation((l) => print('location: $l'));
await Tracelet.start();
// Put the phone on a desk. Wait as long as you like.
// Expected: one fix for the position the session started at.
// Actual: nothing. `Tracelet.getLocations()` stays empty and onLocation never fires
//         until you pick the phone up and walk.
```

Affects `motionDetectionMode` SMART, SPEED and RAW alike, with and without a foreground service.

## Recommended fix

Acquire **one** location at `start()` when the session begins stationary, routed through the normal
location pipeline (filters → Kalman → odometer → persist → `onLocation`), and leave the committed
pace alone. This is what every mature background-geolocation implementation does: a stationary
session still needs a position to anchor itself to.

Explicitly *not* the fix: flipping the default pace to moving. That trades a missing location for a
GPS stream nobody asked for and changes the documented meaning of `motion.isMoving`.

Constraints for the fix:

* Fresh `start()` only. A **resume** (`start(isResume: true)`, the killed-state / relaunch path)
  must keep its current behaviour — it fires on every process relaunch, and its pace is restored
  rather than committed.
* Only when the continuous stream is **not** already running, so the moving branch and the
  `geofenceHighAccuracyMode` branch (#357) are untouched.
* No change to pace transitions, to `changePace()`, or to the geofence subsystem.

GalacticTitan:
## The bug

A continuous session that starts in the **stationary** pace — the default, since
`motion.isMoving` is `false` — acquired **no location at all**. Nothing was delivered until
the accelerometer or activity recognition declared motion, so an app could call `start()`,
put the phone on a desk, and watch `onLocation` stay silent indefinitely.

`start()` branches on the committed pace, and only the moving branch touched the provider.
`changePace(false)` does not fill the gap, and in SMART mode it cannot: `syncCurrentMode()`
sets the coordinator's posture to `STATIONARY_PERIODIC` from the pace just committed, and
both inputs are only then pushed to stationary — so `evaluate_state` sees no mode change,
returns `None`, and never arms the periodic worker either. The one-shot that does exist is
fired from `changePace(true)`, a stationary → moving *transition* that a session beginning
stationary never takes.

## It is a regression

`start()` acquired unconditionally until `bb8af6a0` ("prepare 3.2.0"), which replaced
`locationEngine.start()` with the pace branch — the same edit on both platforms, in one
squashed release commit. The battery intent was right and is kept (later formalised as
#319), but the stream had been doing double duty as the ongoing feed *and* the initial fix,
and only the first job was replaced. Every release since 3.2.0 has started dark.

## The fix

One provider request from the stationary branch (`LocationEngine.requestStartupFix` on both
platforms), routed through the ordinary pipeline — filtered, odometer-counted, persisted
under the configured `persistMode`, dispatched as a normal `location`, and fed to the
geofence evaluator like any other fix.

Deliberately **not** the other option: flipping the default pace to moving buys the same
first location with a full-rate GPS stream nobody asked for, and overrides a documented
config value. The pace the app asked for is left exactly where it was.

## Regression scope

Every path that worked before behaves identically:

| Path | Behaviour |
|---|---|
| Moving start | Untouched — the stream is already acquiring, so the request is skipped (`isTracking` guard) |
| Resume / killed-state relaunch | Untouched — `!isResume` guard; `LocationService`, reconcile, boot receivers, `PeriodicLocationWorker` are not modified |
| `changePace` in either direction | Untouched — the #54 transition one-shot keeps its own trigger and its configured priority |
| Geofences | No geofence file modified; the #357 stationary-stream branch is skipped by the same `isTracking` guard |
| `persistMode` | The anchor goes through `persistLocationIfAllowed`, including the #383 gate |
| Periodic / `getCurrentPosition` | iOS guards on `isPeriodicTracking` and an in-flight sampling window; Android runs after `stopPeriodic()` |
| Autotune / transport classifier (#299) | Unreachable from the anchor — see below |
| Adaptive sampling / elasticity | Unchanged: the anchor is either the first fix ever (distance filter bypassed) or measured against a point far outside any filter |
| Odometer | Same code path and same total; the inter-session distance is added at the anchor instead of at first motion |

## The one real hazard, and how it is closed

Both engines and the Rust processor derive speed from *their own* last fix when the platform
reports none, and `stop()` clears neither reference. So the first fix of a new session in a
live process is derived against wherever the previous session ended: a device carried 5 km
between two sessions yields ~8 m/s — inside `maxImpliedSpeed` (80 m/s) so nothing discards
it, and well above `speedMovingThreshold` (1.5 m/s) so the speed machine wakes on it. That
would overturn the pace `start()` committed a few lines before asking for the anchor: the
silent override of a committed pace that #344 and `StartCommittedPaceTest` exist to prevent.

The anchor therefore contributes only a **measured** speed. A Doppler reading is a real
measurement and still counts, so a device that genuinely is moving when a stationary session
starts is detected on this very fix; only the derivation is dropped, and only for this one
fix. The rule is applied at every consumer of that speed:

* the speed-motion sink (the input that can overturn the pace),
* `lastEffectiveSpeed` — `start()` seeds the machine from it, so an anchor's derived value
  would otherwise reach the machine one session later through that door,
* `rawSpeedSink`, which feeds the transport classifier behind the #299 auto-tune,
* the dispatched location payload.

`motionSpeed` is identical to `result.effectiveSpeed` for every fix that is not the anchor,
so no other fix changes in any way.

On the auto-tune specifically: the classifier is opt-in (`enableFusedClassifier`, default
false), `rawSpeedSink` only stores the latest raw speed, classification runs on the ~1 Hz
accel-window loop, and `applyTransportModeTuning` fires only on a *commit* — confidence-gated
and dwell-debounced (8 s default). One anchor could not commit a mode even before the rule
above; now it cannot bias the stored speed either.

## Verification

* **Android** — `StationaryStartInitialFixTest` (6 tests) drives the wiring end to end: the
  fresh stationary start acquires exactly once, the fix reaches `sendLocation` and
  `lastLocation`, a resume and a moving start acquire nothing, a passive `desiredAccuracy` is
  floored so the request can actually return, and a stale cross-session reference cannot flip
  the committed pace. **Three of the six fail on the pre-fix code** (verified by disabling the
  call and re-running). Full native suite green: 395 tests, 0 failures, stable across three
  consecutive runs.
* The test resets the `TraceletSdk` singleton around each test — Robolectric shares one
  sandbox across every class at the same SDK level, so the engine otherwise belongs to
  whichever class Gradle scheduled first, and the result depended on test order.
* **iOS** — `LocationEngineStartupFixTests` covers the request and its guards (idle,
  already-streaming, unauthorized, and that the #54 transition one-shot still fires
  afterwards), and is **wired into `Package.swift`** — without which it would never compile or
  run, as several stale files in that directory silently do not. Full XCTest suite green on
  the simulator.
* `melos format` and `melos analyze` clean.
* Adds the #385 example card: it starts a session at the default stationary pace and waits
  without moving for a location, then checks the pace was left stationary and that
  `changePace` still round-trips — so the card fails if the fix were ever reintroduced as a
  disguised `isMoving: true`.

Fixes #385


GalacticTitan:
Adjacent finding, filed separately as #387 and deliberately **not** addressed here.

`setOdometer()` writes the total but not the reference it is measured from — the Rust `LocationProcessor` keeps its own `odo_last_*` anchor, which nothing on either platform ever clears (`LocationProcessor.reset()` exists and has no callers outside the core). So "reset to zero, then start tracking" already begins the trip with however far the device travelled while untracked.

That predates this PR and is unchanged by it in amount: for any session that eventually moves, the same leg was previously folded into the first moving fix instead. What this PR changes is *when* it becomes visible on the stationary path — the fix carrying it now arrives at `start()` rather than minutes later — which is how the behaviour was noticed. Kept out of scope because fixing it alters the shared core's odometer contract on all three platforms.

GalacticTitan:
Device run of the #385 card caught a regression this PR introduced, fixed in 75c01e81.

Every row passed except phase 2: after `changePace(true)`, no location in 30s — on both platforms.

The anchor had taken the processor's first-fix slot. `LocationProcessor` waives the distance filter only for a fix with no predecessor:

```rust
if state.last_latitude.is_some() && distance < effective_distance {
    return LocationProcessorResult::filtered("DISTANCE_FILTER", ...);
}
```

Before the anchor existed, the fix that woke a stationary session — the #54 one-shot, or the first fix of the stream the SMART coordinator starts — *was* that first fix and got through for free. With the anchor holding the slot it is a ~0 m duplicate and is dropped, so the app is told it is moving and handed no position. #54's guarantee was withdrawn silently, and on **every** wake path, not just the explicit `changePace` the card exercised — the accelerometer wake had the same hole.

An accepted anchor now arms the force-accept flag, handing back the pass it consumed. Android has carried `forcePersistNextFilteredLocation` since its killed-state wake needed it for exactly this reason; iOS had no equivalent, so `didUpdateLocations` gains the same branch and the same `motionchange` tag. Arming it off the anchor rather than patching `changePace` is what covers the coordinator and foreground-service wakes as well.

New regression test reproduces the card's failure and fails without the hand-back. Android 396 green, full iOS suite green.

PR #417, opened 2026-08-26
Fix/415 permission callback detach

### LINKED ISSUE #415: [Bug]: Android permission request Future is orphaned when the Activity detaches
### Package Name

tracelet (Core Plugin)

### Tracelet Version

3.8.7

### Platform

Android

### Device & OS Version

Pixel 6a, Android 17 / API 37

### Description / Scenario

Disclaimer that this investigation, fix and writeup was done by gpt-5.6-sol low. I have confirmed it resolves the issue in my app.

On a clean installation, `requestLocationAuthorization()` can remain pending after the Android foreground-location permission dialog completes.

`TraceletSdk.requestPermission()` stores the Pigeon reply in `pendingPermissionCallback`. The reply is completed only through `TraceletAndroidPlugin.onRequestPermissionsResult()`.

Both Activity-detach callbacks remove the permission-result listener and clear `sdk.activity`. `onDetachedFromActivity()` does not call the existing `TraceletSdk.clearPendingPermissionCallback()`. If detachment occurs while the permission dialog is active, the only completion path is removed and the Dart Future remains unresolved.

A second request succeeds because `pendingPermissionCallback != null` causes `requestPermission()` to return the current, now-granted authorization synchronously.

Affected lifecycle method:

```text
packages/tracelet_android/android/src/main/kotlin/com/ikolvi/tracelet/flutter/TraceletAndroidPlugin.kt
```

Pending callback implementation:

```text
sdk/android/tracelet-sdk/src/main/kotlin/com/ikolvi/tracelet/sdk/TraceletSdk.kt
```

The cleanup existed previously and was removed during the lifecycle rewrite:

- Added: https://github.com/Ikolvi/Tracelet/commit/afd6f472
- Removed: https://github.com/Ikolvi/Tracelet/commit/b354456c8c712eff2d12eebb573b4081cbfcd635

## 

### Steps to Reproduce / Code Snippet

```dart
final state = await Tracelet.ready(config);

// Clean install: Android displays the foreground-location dialog.
await Tracelet.requestLocationAuthorization()
    .timeout(const Duration(seconds: 10));

// On the affected lifecycle path, this times out despite permission being
// granted in Android settings.
```

Deterministic native regression:

1. Invoke `TraceletHostApiImpl.requestPermission` and capture the unresolved reply.
2. Simulate `onDetachedFromActivity` before `onRequestPermissionsResult`.
3. Assert the Pigeon reply completes exactly once with current authorization.
4. Assert a later permission request is not blocked by stale pending state.

### Tracelet Doctor Bug Report

```markdown

```

### Logs / Stack Traces

```text

```

### Check before submitting

- [x] I have read the Tracelet documentation and FAQs.
- [x] I have searched for existing issues.

mattyg:
Disclaimer this implementation is by gpt-5.6-sol low

Resolves #415

### DIFF
--- a/example/lib/issues/issue_415_card.dart
+++ b/example/lib/issues/issue_415_card.dart
@@ -28,6 +28,8 @@ class _Issue415CardState extends State<Issue415Card>
     final stopwatch = Stopwatch()..start();
 
     try {
+      // The shell's normal preparation is disabled so this card owns the
+      // first permission request on a clean install.
       await Tracelet.ready(
         const Config(
           app: AppConfig(stopOnTerminate: true, startOnBoot: false),
@@ -35,6 +37,7 @@ class _Issue415CardState extends State<Issue415Card>
           http: HttpConfig(autoSync: false),
         ),
       );
+      // Turn an orphaned Pigeon reply into an observable card failure.
       await Tracelet.requestLocationAuthorization().timeout(
         const Duration(seconds: 15),
       );
--- a/example/lib/issues/issue_415_card.dart
+++ b/example/lib/issues/issue_415_card.dart
@@ -1,0 +1,79 @@
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+import 'package:tracelet_example/issues/issue_card_state.dart';
+
+/// Issue #415 — Activity detach could orphan the foreground-permission reply.
+///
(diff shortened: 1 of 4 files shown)


GalacticTitan:
Thanks for the investigation and for confirming the fix in your app — the diagnosis in the linked issue is right, and it's a real regression.

I'm not merging outside PRs into the SDK at the moment, so I'm closing this one and landing the fix myself on the release branch, with the issue credited to you in the changelog. The issue stays open until that ships so you can track it there.

PR #418, opened 2026-08-26
Fix/416 fresh current position

### LINKED ISSUE #416: [Bug]: Android getCurrentPosition retries getCurrentLocation instead of collecting fresh updates
### Package Name

tracelet (Core Plugin)

### Tracelet Version

3.8.7

### Platform

Android

### Device & OS Version

Pixel 6a, Android 17 / API 37

### Description / Scenario

Disclaimer that the investigation and write-up was done by gpt-5.6-sol low. I have confirmed it resolves the issues in my app.

`LocationEngine.collectSamples()` uses repeated `getCurrentLocation()` calls, while the public method comment and the 3.8.x changelog claim that `getCurrentPosition()` uses `requestLocationUpdates()` to force a fresh provider fix.

On the affected device, `getCurrentLocation()` repeatedly returns null or a cached result. Tracelet retries every 800 ms until timeout. Android logs show repeated one-shot registrations without a usable Dart result.

The Play Services wrapper also maps failures to `onSuccess(null)`, hiding the failure reason and feeding the retry loop.

A foreground current-position request should register one update callback, complete from a fresh provider fix, and always remove that callback.

## 

```dart
final location = await Tracelet.getCurrentPosition(
  desiredAccuracy: DesiredAccuracy.high,
  maximumAge: 0,
  samples: 1,
  persist: false,
  timeout: 15,
);
```

Deterministic native regression:

1. Configure `getCurrentLocation()` to return null.
2. Configure `requestLocationUpdates()` to deliver a GPS fix.
3. Call `getCurrentPosition(samples: 1)`.
4. Assert the result completes exactly once from the delivered update.
5. Verify `removeLocationUpdates()` is called.

## Observed Native Signature

```text
getCurrentLocation
getCurrentLocation
getCurrentLocation
...
```

Android provider traces showed repeated immediate one-shot registrations. After the native patch, one continuous update request remained active until a fresh GPS fix arrived and was then removed.

## Expected Behavior

For `samples: 1`, `getCurrentPosition()` must complete from the first fresh provider update. All success, timeout, cancellation, and error paths must remove the update callback.

This behavior was previously reported in comments on
[[issue #46](https://github.com/Ikolvi/Tracelet/issues/46)](https://github.com/Ikolvi/Tracelet/issues/46). The maintainer stated
that v1.8.7 changed `getCurrentPosition(samples: 1)` to use
`requestLocationUpdates()` so it would wake GPS and return a fresh fix.

Tracelet 3.8.7 has regressed: `LocationEngine.collectSamples()` again performs
repeated `getCurrentLocation()` calls. This issue should be described as a
regression of the fix documented in #46.

### Tracelet Doctor Bug Report

```markdown

```

### Logs / Stack Traces

```text

```

### Check before submitting

- [x] I have read the Tracelet documentation and FAQs.
- [x] I have searched for existing issues.

mattyg:
Disclaimer this implementation is by gpt-5.6-sol low.

Resolves #416

### DIFF
--- a/example/lib/issues/issue_416_card.dart
+++ b/example/lib/issues/issue_416_card.dart
@@ -28,6 +28,7 @@ class _Issue416CardState extends State<Issue416Card>
     final stopwatch = Stopwatch()..start();
 
     try {
+      // Permission is a precondition so this card does not depend on #415.
       final authorization = await Tracelet.getLocationAuthorization();
       if (authorization != AuthorizationStatus.whenInUse &&
           authorization != AuthorizationStatus.always) {
@@ -57,6 +58,7 @@ class _Issue416CardState extends State<Issue416Card>
 
       final capturedAt = DateTime.parse(location.timestamp).toUtc();
       final age = DateTime.now().toUtc().difference(capturedAt);
+      // The bound allows delivery latency while rejecting a cached position.
       final fresh = !age.isNegative && age <= const Duration(seconds: 10);
       setStatus(
         fresh
--- a/example/lib/issues/issue_416_card.dart
+++ b/example/lib/issues/issue_416_card.dart
@@ -1,0 +1,100 @@
+import 'package:flutter/material.dart';
+import 'package:tracelet/tracelet.dart' hide State;
+import 'package:tracelet_example/issues/issue_card_shell.dart';
+import 'package:tracelet_example/issues/issue_card_state.dart';
+
+/// Issue #416 — Android current-position sampling regressed to cached one-shots.
+///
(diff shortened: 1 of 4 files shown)


GalacticTitan:
Thanks for the investigation and for confirming the fix in your app — the diagnosis in the linked issue is right, and it's a real regression.

I'm not merging outside PRs into the SDK at the moment, so I'm closing this one and landing the fix myself on the release branch, with the issue credited to you in the changelog. The issue stays open until that ships so you can track it there.

PR #420, opened 2026-08-29
Converge one-shot acquisition on accuracy target

mattyg:
Implements the native contract required by buildyourwebapp/trailhead-wayfinder#1182.

## Problem
`getCurrentPosition(samples: 1)` completed on the first network-derived candidate and shut GPS down within milliseconds. Repeating one-shots never kept GNSS warm long enough to refine.

## Changes
- add `accuracyTarget` and caller-owned `requestId`
- add `cancelCurrentPosition`
- keep Android and iOS providers active until target or caller deadline
- return the best valid observed candidate at timeout
- keep cache fast paths target-aware and cancellable through async enrichment
- ignore invalid Core Location accuracy samples
- add Kotlin, Swift, Pigeon, Dart, and real-device issue-card coverage
- pin FRB codegen 2.12 so generated native and Dart runtimes stay compatible

## Evidence
On the reported Pixel, the old implementation registered HIGH_ACCURACY for 60–200 ms, accepted a network fix, and stopped GPS. The candidate build held the listener continuously for 18 seconds and then promoted a ≤100 m result indoors.

## Validation
- release preflight passed
- bundled Android plugin build passed
- bundled iOS plugin build passed
- Android SDK native tests passed
- iOS SDK tests and example build passed
- Pigeon transport suite passed
- real Trailhead Android app started and converged using locally built fork artifacts

The fork is intended only as a temporary Trailhead dependency until this contract is released upstream.


mattyg:
sorry let me tidy this up

